PR #12, opened 2018-01-29
fix(extensions): overwrite src & srcdoc via cmd args

axe312ger:
## Summary

Easy up developing UI extensions

## Description

If `src` or `srcdoc` args set via CLI, any of these values get removed from the descriptor file to allow overwriting and avoid `Must contain exactly one of: src, srcdoc`

### DIFF
--- a/lib/cmds/extension_cmds/utils/prepare-data.js
+++ b/lib/cmds/extension_cmds/utils/prepare-data.js
@@ -10,6 +10,14 @@ export default async function prepareData (argv) {
   const args = getExtensionData(argv)
   const descriptorArgs = getExtensionData(descriptor)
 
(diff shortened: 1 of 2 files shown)


lib/cmds/extension_cmds/utils/prepare-data.js
```diff
@@ -10,6 +10,11 @@ export default async function prepareData (argv) {
   const args = getExtensionData(argv)
   const descriptorArgs = getExtensionData(descriptor)
 
+  if ('src' in args || 'srcdoc' in args) {
```
Khaledgarbaya: Can you add a comment here explaining the priority of the args over descriptor

PR #15, opened 2018-02-02
Drop unnecessary babel dependencies 

realityking:
## Summary

Drops unnecessary dependencies related to babel
## Description

* `babel-runtime` and `babel-plugin-transform-runtime` can be dropped since the minimum Node.js version is 6 which already supports generators (regenerator support was the only part of `babel-plugin-transform-runtime` that was enabled)
* `babel-types` and `babel-template` aren't used. They're still indirect dependencies,

## Motivation and Context

Less dependencies is better. It doesn't have any impact on the built code.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -7,12 +7,7 @@
     }]
   ],
   "plugins": [
-    "transform-object-rest-spread",
-    ["transform-runtime", {
-      "helpers": false,
-      "polyfill": false,
-      "regenerator": true
-    }]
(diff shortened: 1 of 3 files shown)


axe312ger:
Neat, but I wonder if this breaks the tests. CI for some reason was not running so I enabled it again. Can you please repush this branch to trigger it?

realityking:
Update and Travis passes 🙂

PR #23, opened 2018-04-19
Fixed wrong success message from `space use`

floelhoeffel:


lib/cmds/space_cmds/use.js
```diff
@@ -25,7 +25,7 @@ export const builder = (yargs) => {
 }
 
 function showSuccess (space) {
-  success(`Now using Space ${highlightStyle(space.name)} (${highlightStyle(space.sys.id)}) when the \`--space-id\` option is missing.`)
```
Khaledgarbaya: Hey @floelhoeffel Actually we need that, to let the user know that they override the default space id

floelhoeffel:
@Khaledgarbaya refined - please have another look!

Khaledgarbaya:
@floelhoeffel  we need to update the test snapshots to reflect the new changes. 
please run `npm run test:e2e:run -- -u` and commit the new snapshots.

PR #25, opened 2018-04-20
Add `environment use` sub command

loudmouth:
Adds the ability to store an `"activeEnvironmentId"` in `~/.contentfulrc.json` via `contentful space environment use` with the possibility of either selecting an environment or specifying one with an `--environment-id $ENV_ID` argument.

Also alters the behavior of `contentful space use` to always reset the `"activeEnvironmentId"` to `"master"` and changes the output of the command to reflect this change.

**Still TODO**
- [x] Add tests for `contentful space use`
- [x] Add tests for `contentful space environment use`


loudmouth:
@Khaledgarbaya this PR is ready for review :bam: :captain:

loudmouth:
also realized i forgot to actually read the `activeEnvironmentId` anywhere so that it is used for other commands. i'll have to add that as well.

lib/cmds/space_cmds/environment_cmds/use.js
```diff
@@ -0,0 +1,88 @@
...
+  await assertLoggedIn()
+
+  const { cmaToken, activeSpaceId } = await getContext()
+  const environmentId = argv.environmentId
+
+  const client = await createManagementClient({
+    accessToken: cmaToken
+  })
+
+  const space = await client.getSpace(activeSpaceId)
```
Khaledgarbaya: Sometimes the `activeSpaceId` is not there, we should tell the user instead of showing him a 404 error.
I suggest
1. Add an optional `--space-id` 
2. if `--space-id` is specified you take it and get the environment from there
3. if not you show an error to the user tellig him to pick a space

loudmouth: let's go with 3

PR #32, opened 2018-05-16
[WIP] feat(generator): Add generate subcommand and migration generator

dlitvakb:
## Summary

Adds a `generate` sub-command, and adds a migration generation utility

## Motivation and Context

Creates a migration script of the current content model, either totally or partially, to allow users to properly version their content model.

## Todos

- [x] Implemented feature
- [ ] Add tests

### DIFF
--- a/lib/cmds/space_cmds/generate_cmds/migration.js
+++ b/lib/cmds/space_cmds/generate_cmds/migration.js
@@ -1,0 +1,237 @@
+import { getContext } from '../../../context'
+import { assertLoggedIn, assertSpaceIdProvided, assertEnvironmentIdProvided } from '../../../utils/assertions'
+import { handleAsyncError as handle } from '../../../utils/async'
+import { log, success } from '../../../utils/log'
+
+const recast = require('recast')
+const camelCase = require('camelcase')
+const b = recast.types.builders
+const _ = require('lodash')
+const toAst = require('to-ast')
+const prettier = require('prettier')
+const { createClient } = require('contentful-management')
+const fs = require('fs')
+
+export const command = 'migration'
+
+export const desc = 'Generate a migration file for your content model or a specific content type'
+
+export const builder = (yargs) => {
+  return yargs
+    .usage('Usage: contentful space generate migration')
+    .option('space-id', {
+      describe: 'ID of the space the content model will belong to',
+      type: 'string'
+    })
+    .option('environment-id', {
+      describe: 'ID of the environment the content model will belong to',
+      type: 'string'
+    })
+    .option('content-type-id', {
+      describe: 'ID of the content type to generate. If not provided, will generate the migration for the whole content model in the space',
+      type: 'string'
+    })
+    .option('filename', {
+      describe: 'Output filename. If not provided will generate one with the format SPACE_ID-ENV_ID[-CT_ID]-TIMESTAMP',
+      type: 'string'
+    })
+    .epilog([
+      'See more at:',
+      'https://github.com/contentful/contentful-cli/tree/master/docs/space/generate/migration',
+      'Copyright 2018 Contentful, this is a BETA release'
+    ].join('\n'))
+}
+
+export const aliases = ['m']
+
+const ctVariableEscape = function (ctId) {
+  const camelCased = camelCase(ctId)
+  return !isNaN(ctId.charAt(0)) ? `_${camelCased}` : camelCased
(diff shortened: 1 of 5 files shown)


lib/cmds/space_cmds/generate_cmds/migration.js
```diff
@@ -0,0 +1,237 @@
+import { getContext } from '../../../context'
+import { assertLoggedIn, assertSpaceIdProvided, assertEnvironmentIdProvided } from '../../../utils/assertions'
+import { handleAsyncError as handle } from '../../../utils/async'
+import { log, success } from '../../../utils/log'
+
+const recast = require('recast')
```
Khaledgarbaya: can you change this to `import recast from 'recast'` to keep the same style, also from the other dependencies

lib/cmds/space_cmds/generate_cmds/migration.js
```diff
@@ -0,0 +1,237 @@
...
+const generateFileName = function (spaceId, environmentId, contentTypeId) {
+  const contentTypePart = contentTypeId ? `-${contentTypeId}` : ''
+
+  return `${spaceId}-${environmentId}${contentTypePart}-${Date.now()}.js`
+}
+
+export async function generateMigration (argv) {
+  await assertLoggedIn()
+  await assertSpaceIdProvided(argv)
+  await assertEnvironmentIdProvided(argv)
```
Khaledgarbaya: IMO we can remove the `assertEnvironmentIdProvided`.
for the environment case, we can default it to `master` if it's not provided like the import and export is doing. This will provide a better experience.

lib/cmds/space_cmds/generate_cmds/migration.js
```diff
@@ -0,0 +1,237 @@
...
+    editorInterfacesCreators = contentType.fields.map((field) => {
+      const control = _.find(editorInterface.controls, (control) => {
+        return control.fieldId === field.id
+      })
+      const widgetId = control.widgetId
+      const settings = control.settings
+
+      return b.expressionStatement(changeEditorInterface(contentType.sys.id, field.id, widgetId, settings))
+    })
+  } catch (err) {
```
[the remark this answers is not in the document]
dlitvakb: Not really, if it 404s, it's because there are no fields on the CT, which is an OK state for me, maybe what I can do is check wether or not the CT actually doesn't have fields and act based on that? WDYT?

lib/cmds/space_cmds/generate_cmds/migration.js
```diff
@@ -0,0 +1,237 @@
...
+export const desc = 'Generate a migration file for your content model or a specific content type'
+
+export const builder = (yargs) => {
+  return yargs
+    .usage('Usage: contentful space generate migration')
+    .option('space-id', {
+      describe: 'ID of the space the content model will belong to',
+      type: 'string'
+    })
+    .option('environment-id', {
```
Khaledgarbaya: you can add `default: 'master'` to the option so when the user does not pass an environment the value will default to `master`

lib/cmds/space_cmds/generate_cmds/migration.js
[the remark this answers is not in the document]
Khaledgarbaya: since we are logging `Fetching editor interface` in line 176 we should log in here also to mark the end of this task.
Something like the contentType does not have any fields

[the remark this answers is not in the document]
Roblinde: Yeah, I'm thinking just a log "The content type X does not have any fields".

Just to make sure that when someone is trying to figure out what's going on it's not just swallowed in a catch. I know it's a very rare edge case, but a simple log costs nothing, right.

lib/cmds/space_cmds/generate_cmds/migration.js
```diff
@@ -0,0 +1,275 @@
...
+    'protected', 'public', 'static',
+    'await', 'abstract', 'boolean',
+    'byte', 'char', 'double',
+    'final', 'float', 'goto',
+    'int', 'long', 'native',
+    'short', 'synchronized', 'throws',
+    'transient', 'volatile'
+  ]
+
+  if (~reservedWords.indexOf(ctId)) {
```
Khaledgarbaya: `if(reservedWords.indexOf(ctId) !== -1)`

PR #45, opened 2018-06-12
build(npm): switch to maintained cli-table3

axe312ger:
cli-table2 is no more maintained --> https://github.com/DanielRuf/cli-table3


Khaledgarbaya:
What about this https://github.com/contentful/contentful-cli/pull/44

Khaledgarbaya:
@axe312ger ci is failing

package.json
```diff
@@ -37,70 +37,70 @@
...
+    "contentful-management": "^5.1.4",
+    "contentful-migration": "^0.9.0",
     "emojic": "^1.1.11",
-    "execa": "^0.8.0",
+    "execa": "^0.10.0",
     "figlet": "^1.2.0",
-    "https-proxy-agent": "^2.0.0",
-    "inquirer": "^3.2.3",
+    "https-proxy-agent": "^2.2.1",
+    "inquirer": "^6.0.0",
```
axe312ger: might fix OAUTH windows issue with copy & paste

lib/cli.js
```diff
@@ -17,14 +25,18 @@ yargs.usage('\nUsage: contentful <cmd> [args]')
     describe: 'Show current version',
     skipValidation: true
   })
+  .version(false)
```
axe312ger: We have our own version implementation to avoid conflict with `contentful space extension update --version`. yargs 9 did enable this by default, so I had to disable it again.

lib/cli.js
```diff
@@ -2,6 +2,14 @@ import yargs from 'yargs'
...
 import {version} from '../package.json'
 
+// Workaround to remove display of script name in help. See:
+// * https://github.com/yargs/yargs/pull/1143
+// * https://github.com/yargs/yargs/issues/1065
+// * https://github.com/yargs/yargs/issues/1048
+function fixHelp (output) {
+  return output.replace(/( +)contentful\.js/g, '$1')
+}
+
```
axe312ger: Nasty ugly bug adding `contentful.js` to our help texts. This hack removes that for now, should be fixed soonish by the community ™️

PR #47, opened 2018-06-18
Add management-token

Khaledgarbaya:
This PR allow users to pass `management-token` when using the cli.

TODO: 
- [x] Check the rest of the command that requires a `management-token`
- [x] Update Documentations


lib/utils/assertions.js
```diff
@@ -2,14 +2,14 @@ import { getContext } from '../context'
 import { PreconditionFailedError } from './error'
 import { highlightStyle } from './styles'
 
-export async function assertLoggedIn () {
+export async function assertLoggedIn ({managementToken} = {}) {
   const context = await getContext()
-  if (!context.cmaToken) {
+  if (!context.cmaToken && !managementToken) {
     throw new PreconditionFailedError(`You have to be logged in to do this. You can log in via ${highlightStyle('contentful login')}`)
```
loudmouth: I think the error message change here now. something like `You have to be logged in, or provide a management token argument to do this.` Not sure on exact wording.

PR #56, opened 2018-07-24
feat(Context): Allow config from local folder

### LINKED ISSUE #43: Allow config and running from local project folder
## Expected Behavior

Having the option to install your config locally within a project root folder (`~/path/to/project/.contentfulrc.json`) rather than the OS user home folder (`~/.contenfulrc.json`). 

This also means being able to run contentful commands, for example, through NPM and have the nearest config referenced, e.g. `~/path/to/project/.contentfulrc.json`.

## Actual Behavior

Config is installed in OS user home folder (`~/.contenfulrc.json`), making different configs for different projects a bit of a messy affair.

phoebeschmidt:
Allow contentfulrc config to live anywhere, not just home directory.

fixes #43

## Summary

Use [find-up](https://github.com/sindresorhus/find-up) to search up file tree `contentfulrc.json`. If none is found, use home directory: `~/.contentfulrc.json`


lib/context.js
```diff
@@ -1,13 +1,17 @@
 import { readFile, writeFile } from 'mz/fs'
 import { resolve } from 'path'
 import { homedir } from 'os'
+import { sync } from 'find-up'
```
axe312ger: sync might cause confusion here. Better import the whole `find-up` and call `findUp.sync`

lib/context.js
```diff
@@ -1,13 +1,17 @@
...
 
 import { proxyStringToObject } from './utils/proxy'
 
 let context
 
 export function getConfigPath () {
-  return resolve(homedir(), '.contentfulrc.json')
+  const contentfulrc = '.contentfulrc.json'
+  const configPath = sync(contentfulrc)
```
axe312ger: Did you investigate if we can make all the related functions async? I know I said we could do it sync or even have to in current state but I'd prefer using the async apis of our dependencies cus they are in theory more performant.

lib/context.js
```diff
@@ -1,13 +1,17 @@
...
+  if (!configPath) return resolve(homedir(), contentfulrc)
```
axe312ger: Let's add some curly brackets here {}

While it might be okay here I prefer to have them. It is more clear to read and later changes won't cause that big git diffs.

Pure coding style pickiness here 🙈

axe312ger: Using async would even force you to do it in a try-catch block which would then resolve this issue too.

test/unit/context.test.js
```diff
@@ -35,8 +36,16 @@ test.afterEach((t) => {
 })
 
 test('locates correct config file path', (t) => {
+  const configPath = getConfigPath()
+  t.is(configPath, resolve(sync('.contentfulrc.json')))
```
axe312ger: U have a `.contentfulrc.json` on your disc or even your home directory. Fresh installs and the CLI probably won't have that. You may want to add a `.contentfulrc.json` file to the repository root directory while the tests are running. You gonna need this for e2e tests, too.

lib/context.js
phoebeschmidt: i was thinking this 👍

lib/context.js
phoebeschmidt: yea there are async apis!

lib/context.js
Khaledgarbaya: or `import { sync as findUpSync } from 'find-up'`

lib/context.js
```diff
@@ -1,13 +1,23 @@
...
 let context
 
-export function getConfigPath () {
-  return resolve(homedir(), '.contentfulrc.json')
+export async function getConfigPath () {
+  const contentfulrc = '.contentfulrc.json'
+  try {
+    const configPath = await findUp(contentfulrc)
+    if (!configPath) {
+      throw Error('no config path found')
```
Khaledgarbaya: Is there any reason we changed this to throw if we don't find the path, we can maybe send back `resolve(homedir(), '.contentfulrc.json')` if we don't find anything

Khaledgarbaya: One other thing maybe we can cache the path once we found it, at lease while the process is running.

phoebeschmidt: hm good q. I guess I was trying to cover the case where (1) findUp throws an error and (2) findup finds nothing (aka resolves with `null`) both in the same catch statement.

I'll follow up about caching in person

phoebeschmidt: we chatted in person all resolved 👍

lib/cmds/login.js
```diff
@@ -22,7 +22,7 @@ export async function login () {
 
   if (cmaToken) {
     log()
-    log(`Looks like you already stored a CMA token on your system. ${chalk.dim(`(Located at ${getConfigPath()})`)}`)
+    log(`Looks like you already stored a CMA token on your system. ${chalk.dim(`(Located at ${await getConfigPath()})`)}`)
```
axe312ger: this works? 😱

lib/context.js
```diff
@@ -72,13 +82,13 @@ async function loadRuntimeConfig () {
   return context
 }
 
-export function storeRuntimeConfig () {
+export async function storeRuntimeConfig () {
```
axe312ger: Did you check if there are any `storeRuntimeConfig()` calls which do not use await?

lib/cmds/login.js
phoebeschmidt: yes! i tested this 😎

lib/context.js
phoebeschmidt: yep, they all use await. I just had to add one in a test that was broken

PR #59, opened 2018-07-30
Feat/gatsby guide tests

phoebeschmidt:
Unit tests and integration tests for guide command.

- [ ] delete integration test directory after test is over (need help with this)


phoebeschmidt:
@axe312ger fyi: 

Fixed all tests, but CI is still failing because of a 403 from the seed step in the `guide` integration test. Something to do with github permissions maybe?

test/unit/cmds/login.test.js
```diff
@@ -6,74 +6,68 @@ import {
   handler as loginHandler,
```
phoebeschmidt: these tests were failing (but only on TravisCI and not locally 🙈) because of some changes in `login.js` related to `process.platform` etc. I did some pretty major refactoring here to stub more things directly in login.js (and not context.js)

PR #60, opened 2018-08-02
Reworking the tests for better DX

axe312ger:
It is not sure what we will do, I just use this PR for now to gather some smaller fixes & ideas.


axe312ger:
I was hoping this would do some output while the guide test is running. But it waits till the test is done and then outputs all the information. Closing this one since it does not improve developer experience while testing.

PR #66, opened 2018-08-08
Make logged-in assertion check in command parameters

dborsatto:
After the `--management-token` parameter was added to the `space list` command, the `assertLoggedIn` function wasn't updated to receive the `argv` parameter, so it will always fail if no default token is found in the `.contentfulrc.json` file

### DIFF
--- a/lib/cmds/boilerplate.js
+++ b/lib/cmds/boilerplate.js
@@ -46,7 +46,7 @@ async function getBoilerplates () {
 }
 
 export async function downloadBoilerplate (argv) {
-  await assertLoggedIn()
+  await assertLoggedIn(argv)
   await assertSpaceIdProvided(argv)
   const { spaceId } = await normalizer(argv)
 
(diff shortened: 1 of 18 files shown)


lucasvanlierop:
Hi, great to see you're working on this. I just noticed `argv` isn't passed to `assertLoggedIn` in some other places as well.

I also noticed migrations expects `--access-token` (instead of `--management-token`), is this intentional?
https://github.com/contentful/contentful-cli/blob/3897ced4ccb6ecbc7d5c6ebcebd9118422484a86/lib/cmds/space_cmds/migration.js#L56

I also noticed the default value of the `paramName` hints at `--management-Token` instead of `--management-token`
https://github.com/contentful/contentful-cli/blob/3897ced4ccb6ecbc7d5c6ebcebd9118422484a86/lib/utils/assertions.js#L7

Khaledgarbaya:
> I also noticed the default value of the paramName hints at --management-Token instead of --
management-token

that's a typo yes but it does not affect the code.

> I also noticed migrations expects --access-token (instead of --management-token), is this intentional?

That's some legacy naming coming from the migration tool and we'll be aligned by `v1`

> I just noticed argv isn't passed to assertLoggedIn in some other places as well.

maybe @dborsatto can check these and fix them in this PR

Khaledgarbaya:
available on `v0.16.1`

PR #73, opened 2018-08-21
fix(contentful-clients.js): send proxy object to CMA client

phoebeschmidt:
## Summary
Discovered a bug when implementing proxy to record/replay http requests in integration tests. Simply sending the `proxy` object instead of creating an `httpsAgent` instance fixes the problem. Wanted to discuss if this solution is viable.

### DIFF
--- a/lib/utils/contentful-clients.js
+++ b/lib/utils/contentful-clients.js
@@ -1,14 +1,12 @@
 import { createClient } from 'contentful-management'
 import { version } from '../../package.json'
 import { getContext } from '../context'
-import { agentFromProxy } from './proxy'
 
 export async function createManagementClient (params) {
   params.application = `contentful.cli/${version}`
 
   const context = await getContext()
   const { proxy, host = 'api.contentful.com' } = context
-  const { httpsAgent } = agentFromProxy(proxy)
 
-  return createClient({ ...params, httpsAgent, host })
+  return createClient({ ...params, proxy, host })
 }


Khaledgarbaya:
I don't think that's the case

zcei:
I'm afraid this passing with `proxy` is a false positive.

We carefully created this config and tested that against enterprise setups that were failing beforehand (when the proxy setting was available). 
A test suite replaying some requests is not really good to test e.g. HTTPS termination, which is a common theme. We had cases of HTTPS via HTTP proxies or HTTP via HTTPS proxies.

axe312ger:
http proxy over httpsAgent was what we did not get to work, what configuration did you use? 🤔

axe312ger:
@zcei btw.: how can it be false positive when our replay proxy is able to record the requests? It is working very well when axios itself creates the proxy connection 🤔

zcei:
Yeah, this was what we figured as well - some proxy combinations worked, we thought everything was fixed and still it yielded problems. I don't know what kind of replay proxy you're using, but the one I initially wrote for the migration project (I think that's the one here in the CLI) wasn't too reliable. Just one setup to satisfy that hardly reflects real world use-cases.

I sound very grumpy, I know - that's just because I have no clue how to better test it without spawning some docker-compose setups that contain real proxies. 🤷‍♂️

axe312ger:
@zcei yeah, thats what I am going to propose today in the standup. Keep u posted.

PR #77, opened 2018-08-23
end 2 end test on windows, linux & osx

axe312ger:
run simple e2e tests with the standalone CLI

* appveyor for windows
* circleci for linux & osx


Khaledgarbaya:
Hey @axe312ger what do you think of limiting this just for master ?

PR #80, opened 2018-08-28
feat(proxy): add new raw proxy config option

axe312ger:
Depends on:
* https://github.com/contentful/contentful-import/pull/113
* https://github.com/contentful/contentful-export/pull/118
* https://github.com/contentful/contentful-migration/pull/130

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,15 @@ When multiple proxy configurations exists, precedence is taken in this form:
 2. `https_proxy` takes precedence over `.contentfulrc.json`
 2. `https_proxy` takes precedence over `http_proxy`
 
+
+## :rescue_worker_helmet: Troubleshooting
+
+* Unable to connect to Contentful through your Proxy? Try settings `rawProxy: true` in your `.contentfulrc.json` via:
+
+```sh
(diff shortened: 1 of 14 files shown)


README.md
```diff
@@ -75,6 +75,14 @@ When multiple proxy configurations exists, precedence is taken in this form:
 2. `https_proxy` takes precedence over `.contentfulrc.json`
 2. `https_proxy` takes precedence over `http_proxy`
 
+### Proxy troubleshooting
+
+* Unable to connect? Try settings `rawProxy: true` in your `.contentfulrc.json` via:
+
+```sh
+contentful config add --raw-proxy true
```
Khaledgarbaya: `--raw-proxy` is enough since this is a boolean

axe312ger: No it is not, the command will complain otherwise.

```
$ ./bin/contentful.js config add --raw-proxy
/...contentful-cli/dist/cli.js:28
  if (err) throw err;
           ^
YError: Not enough arguments following: raw-proxy
...
```

axe312ger: I removed the part from the code now which requires the true. WIll see in my next commit.

PR #82, opened 2018-08-29
Test/proxy recording

phoebeschmidt:
## Summary

Introducing talkback proxy to record/replay http requests in integration tests.

All the `recordings/*` files can be ignored. Focus on new pretest/posttest scripts, updated npm script names, and changes to existing tests.

Blocked by https://github.com/contentful/contentful-management.js/pull/187

TODO
- [x] update pre- and post- test scripts to find config file in tree (it might not necessarily live at the root)


phoebeschmidt:
also every integration test is timing out on travis, so I'm assuming the talkback proxy isnt configured correctly there. Will need some assistance with this

package.json
```diff
@@ -24,27 +24,29 @@
...
-    "test:integration": "npm run test:integration:setup && npm run test:integration:nosetup && npm run config:restore",
-    "test:integration:ci": "npm run test:integration:setup && npm run test:integration:nosetup:ci && npm run config:restore",
-    "test:integration:fresh": "npm run test:integration:setup && npm run test:integration:update && npm run config:restore",
-    "test:integration:setup": "test/integration/scripts/pre-test.sh",
-    "test:integration:nosetup": "BABEL_ENV=test jest test/integration/cmds/** --verbose",
-    "test:integration:nosetup:ci": "BABEL_ENV=test jest test/integration/cmds/** --verbose --runInBand",
-    "test:integration:update": "BABEL_ENV=test jest test/integration/cmds/** --verbose --updateSnapshot",
-    "test:integration:watch": "BABEL_ENV=test jest test/integration/cmds/** --verbose --watch",
+    "pretest:integration": "node test/integration/scripts/pre-test.js",
+    "pretest:integration:ci": "npm run pretest:integration",
```
phoebeschmidt: the duplication in these scripts is not ideal, but this is the best I could come up with so that we could pass the `--runInBand` flag to the integration tests on CI only. running tests without that flag takes less than half the time, so I don't want to enforce it for local tests if we don't have to.

any ideas for improving this are welcome!

test/integration/scripts/post-test.js
```diff
@@ -0,0 +1,26 @@
...
+    console.log('Restoring config file...')
+    await fs.writeFile(configPath, configContent)
+  } catch (e) {
+    if (e.code === 'ENOENT') {
+      // no backup existed, which means no config existed before the test? TODO what to do in this situation
+      console.log('No backup, removing proxies from config...')
+      const configContents = JSON.parse(await fs.readFile(configPath))
+      delete configContents.proxy
+      delete configContents.rawProxy
+      await fs.writeFile(configPath, JSON.stringify(configContents))
```
phoebeschmidt: wasn't sure if this was correct thing to do here.

if there is no backup, that means there was no config file before running the tests. should we just remove the file all together? this felt weird in case these scripts are run solo, you could accidentally delete your config.

phoebeschmidt: also this assumes config lives at root, which is not necesarily the case.  I'll add a to-do to fix this

package.json
Khaledgarbaya: 👍

Khaledgarbaya: improving the workflow is an ongoing thing as long as we update the docs everytime, I think it should be fine

test/integration/scripts/post-test.js
Khaledgarbaya: It would be nice to create a ticket internally also for this

recordings/5-get-environment.json5
```diff
@@ -0,0 +1,122 @@
...
+            'x-permitted-cross-domain-policies': [
+                'none',
+            ],
+            'x-xss-protection': [
+                '1; mode=block',
+            ],
+            connection: [
+                'Close',
+            ],
+            'set-cookie': [
```
Khaledgarbaya: Maybe we should ignore this because we might end up recording sensitive data

phoebeschmidt: decided we dont need to remove this yet

PR #83, opened 2018-08-30
Fix/space create promt

Khaledgarbaya:
This PR adds a confirmation step when creating space to remind users that a new space creation can lead to a monthly charge depends on the plan they have

### DIFF
--- a/lib/cmds/space_cmds/create.js
+++ b/lib/cmds/space_cmds/create.js
@@ -7,6 +7,8 @@ import { assertLoggedIn } from '../../utils/assertions'
 import { EventSystem } from '../../core/events'
 import { CREATE_SPACE_HANDLER } from '../../core/events/scopes'
 
+import { warningStyle } from '../../utils/styles'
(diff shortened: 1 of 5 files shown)


lib/cmds/space_cmds/create.js
```diff
@@ -33,6 +35,11 @@ export const builder = (yargs) => {
       alias: 'org',
       describe: 'Organization owning the new space'
     })
+    .option('yes', {
+      alias: 'y',
```
loudmouth: what do you think about `--force` `-f` instead?

phoebeschmidt: I like `--force` better than `--yes` 👍

Khaledgarbaya: `--yes` is a Linux standardized thing for yes or no questions and we are doing this for other commands

phoebeschmidt: ah okay, didn't realize. then sure, we keep the `--yes` :)

PR #124, opened 2019-02-21
feat(cmds): unify management token and space id params across all commands

phoebeschmidt:
Previously not all commands accepted `--management-token` or `--space-id` as an arg. For this PR i went through each command and made sure it accepted management-token and space-id with the correct alias.

Started to try to make a test that could be reused between all the commands but it became very complex very quickly. It also felt extreme to try to create a custom test for each command. So this is the minimum viable PR to try to make things consistent.


test/integration/cmds/extension/__snapshots__/create.test.js.snap
```diff
@@ -9,7 +9,7 @@ Options:
   -h, --help                 Show help                                 [boolean]
   --id                       Extension id                               [string]
   --name                     Extension name                             [string]
-  --management-token         Contentful management API token            [string]
+  --management-token, --mt   Contentful management API token            [string]
```
Khaledgarbaya: If I remember correctly for the alias it should be `-mt` with one `-`

phoebeschmidt: these are generated by yargs. It seems any multi-letter alias has two dashes and any single letter alias has one?

Khaledgarbaya: 💡

PR #129, opened 2019-03-13
Address deprecated function warning when listing content types

makinwab:
## Summary
Be more Environment aware.

## Description
- Get content types via the environment vs space
- Add new option, environment id, to specify the environment when listing content types
- Update content type list cmd documentation with new options

## Motivation and Context
- Given the environment driven context, certain methods would only be available by an environment object hence deprecation warnings
- Valuable to have context on which environment is active and which one is currently on when listing content types.
- Nice to have, being able to list content types from a specific environment without having to switch to that environment.

### DIFF
--- a/lib/cmds/content-type_cmds/list.js
+++ b/lib/cmds/content-type_cmds/list.js
@@ -47,7 +47,9 @@ async function ctList (argv) {
 
   const space = await client.getSpace(spaceId)
   const environment = await space.getEnvironment(environmentId)
+
   log(highlightStyle(`Environment: "${environmentId}"`))
+
   const result = await paginate({
     client: environment,
     method: 'getContentTypes',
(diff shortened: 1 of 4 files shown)


lib/cmds/content-type_cmds/list.js
```diff
@@ -17,25 +18,33 @@ export const builder = (yargs) => {
...
 
 async function ctList (argv) {
   await assertLoggedIn(argv)
   await assertSpaceIdProvided(argv)
 
-  const { cmaToken, activeSpaceId } = await getContext()
+  const { cmaToken, activeSpaceId, activeEnvironmentId } = await getContext()
   const managementToken = argv.managementToken || cmaToken
   const spaceId = argv.spaceId || activeSpaceId
+  const environmentId = argv.environmentId || activeEnvironmentId
```
Khaledgarbaya: What do you think about setting the `environmentId` to `master` when the user does not pass anything and there is no activeEnvironmentId?

makinwab: Yeah, thought to do this but held back because I didn't know how our conversations regarding implementing environment branching/recovering with aliases was going to affect this. I'll set "master" as a fallback here for now and we can maybe adjust accordingly in the future.

PR #130, opened 2019-03-21
Add --quiet option for migrations

makinwab:
## Summary

Allow for migrations to happen with a quiet mode

## Description

- Add option `--quiet, -q` to the usage params

## Motivation and Context
The [Contentful Migration](https://github.com/contentful/contentful-migration) package now supports a quiet mode. Hence, Contentful Cli needs to be updated accordingly.

## Todos

-   [x] Implemented feature

Screen Shot 2019-03-19 at 11 42 42


phoebeschmidt:
We also have semantic release so just make sure you squash/merge and that the commit message is sem-rel format 👍

PR #133, opened 2019-03-25
feat(config): change config look up order

suevalov:
## Summary

This PR changes `.contentfulrc.json` look up order. 

## Description

*Current behavior*: `.contentfulrc.json` located in `home` folder trumps config located in `process.cwd()`

*New behavior*: script tries to find `.contentfulrc.json` in `process.cwd()` (and then looks one folder up in a tree, recursively). if there's no `.contentfulrc.json` in a tree, then `home` folder is used.

## Motivation and Context

It's convenient to collocate config with a project you're currently working on and be able to jump between projects without a need to change configuration in global configuration file.

Current behavior complicates the flow for extensions development using [create-contentful-extension](https://github.com/contentful/create-contentful-extension).

### DIFF
--- a/lib/context.js
+++ b/lib/context.js
@@ -11,10 +11,8 @@ let configPath
 export async function getConfigPath () {
   const contentfulrc = '.contentfulrc.json'
   const defaultPath = resolve(homedir(), contentfulrc)
-  if (!configPath) {
-    const nestedConfigPath = await findUp(contentfulrc)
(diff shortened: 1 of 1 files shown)


suevalov:
@Khaledgarbaya The interesting thing that current tests cover the behavior. I'd say that they didn't properly cover previous implementation.

```
// context.test.js

test('locates correct config file path')
test('uses home directory as config file path if none is found')
```

PR #150, opened 2019-06-26
fix: Use set environment in commands

### LINKED ISSUE #68: Command "space migration" ignores active environment from .contentfulrc.json
## Expected Behavior

When doing:
```
$ contentful space environment use -e "some-env"
$ contentful space migration some-script.js
```
I would expect migration to be run against "some-env" environment.

## Actual Behavior

The confirmation dialog shows active environment is ignored:
```
The following migration has been planned

Environment: master

Create Content Type testContent
  - name: "Test content"
  - description: ""
```

## Possible Solution

Consistently support space and environment set in `.contentfulrc.json` to act as default space and environment parameters.

## Environment
- **Node Version:** v8.11.3
- **Npm Version:** 6.1.0
- **Package Version:** 0.16.0

jroehl:
## Summary

Fix for #68 
Find and add `management-token`, `space-id`, `environment-id` and `host` to context.
 
## Description

Added middleware function (using [yargs middlewares](https://github.com/yargs/yargs/blob/master/docs/api.md#middlewarecallbacks-applybeforevalidation)) to build the context by checking if the `management-token`, `space-id`, `environment-id` and `host` are set in the `.contentfulrc.json` or passed into the command via flags.

## Motivation and Context

This fixes #68 and simplifies checking for cli settings.

## Todos

- [x] Implemented feature
- [x] Fix tests


lib/cmds/config_cmds/remove.js
```diff
@@ -61,9 +58,18 @@ export const removeHandler = async (argv) => {
     if (argv.activeSpaceId) {
       delete options['activeSpaceId']
     }
+    if (argv.activeEnvironmentId) {
```
Khaledgarbaya: Maybe we can have something like 
```js
const contextKeys = ['activeEnvironmentId', 'proxy', 'host', 'rawProxy']
contextKeys.forEach(key => {
	if(argv[key]) {
      delete options[key]
	}
})
```
This way when we add support for a new config property we only update the `contextKeys` array

lib/utils/middlewares.js
```diff
@@ -0,0 +1,34 @@
+import {getContext} from '../context'
+
+export const buildContext = async (argv) => {
+  const {
+    managementToken,
+    cmaToken,
```
Khaledgarbaya: I think we should align all arguments to be `managementToken` instead of having 2 property for the same value

lib/cmds/config_cmds/remove.js
jroehl: 👍

lib/utils/middlewares.js
jroehl: Good idea, i changed `cmaToken` to `managementToken` and `cdaToken` to `deliveryToken` to have a consistent naming

lib/cmds/space_cmds/use.js
```diff
@@ -24,32 +24,31 @@ export const builder = (yargs) => {
     .epilog('Copyright 2018 Contentful, this is a BETA release')
 }
 
-function showSuccess (space) {
-  success(`Now using the 'master' Environment of Space ${highlightStyle(space.name)} (${highlightStyle(space.sys.id)}) when the \`--environment-id\` option is missing.`)
+function showSuccess (space, env) {
+  success(`Now using the '${env}' Environment of Space ${highlightStyle(space.name)} (${highlightStyle(space.sys.id)}) when the \`--environment-id\` option is missing.`)
```
phoebeschmidt: shouldn't have single quotes around it, right? otherwise the string interpolation won't work?

jroehl: I would think this still works, as the quotes (single or double) are handled as normal strings within backticks .. Am i mistaken ?

phoebeschmidt: yeah you're right my bad 👍

lib/cmds/extension_cmds/delete.js
```diff
@@ -29,27 +28,26 @@ export const builder = (yargs) => {
...
-  await assertLoggedIn(argv)
-  await assertSpaceIdProvided(argv)
+  const {context, id} = argv
+  await assertLoggedIn(context)
+  await assertSpaceIdProvided(context)
 
-  const { cmaToken, activeSpaceId, activeEnvironmentId } = await getContext()
-  const spaceId = argv.spaceId || activeSpaceId
-  const environmentId = argv.environmentId || activeEnvironmentId || 'master'
+  const { managementToken, activeSpaceId, activeEnvironmentId = 'master' } = context
```
Khaledgarbaya: I think we can move setting of the activeEnvironmentId to `master` to the middlware

jroehl: ✅done 🙂

PR #154, opened 2019-06-28
fix: import --no-update flag

### LINKED ISSUE #38: import --no-update option doesn't work
```
contentful space import --no-update --content-file output.json
Usage: contentful space import --content-file <file>

Options:
  --environment-id           ID the environment in the destination space
                                                    [string] [default: "master"]
  --content-file             JSON file that contains data to be import to your
                             space                           [string] [required]
  --content-model-only       Import only content types[boolean] [default: false]
  --skip-content-model       Skip importing content types and locales
                                                      [boolean] [default: false]
  --no-update                Skips updating entries if they already exist
                                                      [boolean] [default: false]
  --management-host          Management API host
                                        [string] [default: "api.contentful.com"]
  --proxy                    Proxy configuration in HTTP auth format:
                             [http|https]://host:port or
                             [http|https]://user:password@host:port     [string]
  --config                   An optional configuration JSON file containing all
                             the options for a single run

Copyright 2018 Contentful, this is a BETA release

Unknown argument: update
```

Version 0.14.1 on Mac

jroehl:
## Summary

Fixes #38 
Update handling of `--(no-)update` flag according to yargs specification.

## Description

Changed the flag `--no-update` to `--update` defaulting to true. In that way it is possible to specify `--no-update` or `--update false` to disable update of an entry if it exists.

## Motivation and Context

This fixes #38 

## Todos

- [x] Implemented feature
- [x] Fix tests


phoebeschmidt:
not sure it actually totally fixes #38 since this flag isnt actually implemented in https://github.com/contentful/contentful-import 😬. it does fix conventions. maybe we should open an issue on contentful-import if we are going to close #38?

jroehl:
That's true - it's not implemented. But `noUpdate` is, as far as I can tell. That's why I inverted the flag (https://github.com/contentful/contentful-cli/pull/154/files#diff-e0e0c0988ae5dd0c4eb60a59b371e42eR30).

But it would actually be nicer if we wouldn't need to do that. 👍

jroehl:
Apparently it is not yet implemented in `contentful-import`, yet 👉 https://github.com/contentful/contentful-import/issues/118. Thanks @phoebeschmidt pointing that out.

PR #172, opened 2019-08-22
feat: Add Environment aliases management

Khaledgarbaya:
## Description
Add Environment aliases management 

## Todos
- [x] Update the management SDK
- [x] Add Tests

Screenshot 2019-08-28 at 14 38 56

### DIFF
--- a/lib/cmds/space_cmds/alias_cmds/list.js
+++ b/lib/cmds/space_cmds/alias_cmds/list.js
@@ -46,7 +46,7 @@ export async function environmentAliasList ({ context }) {
   const aliases = result.items.sort((a, b) => a.name.localeCompare(b.name))
 
   const table = new Table({
-    head: ['Alias id', 'Target']
+    head: ['Alias id', 'Target environment ID']
   })
   aliases.forEach((alias) => {
     table.push([
--- a/lib/cmds/space_cmds/alias_cmds/list.js
+++ b/lib/cmds/space_cmds/alias_cmds/list.js
@@ -31,7 +31,7 @@ export const builder = (yargs) => {
 
 export const aliases = ['ls']
 
-export async function aliasList ({ context }) {
+export async function environmentAliasList ({ context }) {
   const { managementToken, activeSpaceId } = context
 
   const client = await createManagementClient({
@@ -41,14 +41,13 @@ export async function aliasList ({ context }) {
 
   const space = await client.getSpace(activeSpaceId)
 
-  const result = await paginate({ client: space, method: 'getAliases' })
+  const result = await paginate({ client: space, method: 'getEnvironmentAliases' })
 
   const aliases = result.items.sort((a, b) => a.name.localeCompare(b.name))
 
   const table = new Table({
     head: ['Alias id', 'Aliased to']
   })
-
   aliases.forEach((alias) => {
     table.push([
       alias.sys.id,
@@ -59,4 +58,4 @@ export async function aliasList ({ context }) {
   log(table.toString())
 }
 
-export const handler = handle(aliasList)
+export const handler = handle(environmentAliasList)
--- a/lib/cmds/space_cmds/alias_cmds/list.js
+++ b/lib/cmds/space_cmds/alias_cmds/list.js
@@ -25,7 +25,7 @@ export const builder = (yargs) => {
     })
     .epilog([
       'Environment aliases is an [ALPHA] Feature',
-      'Copyright 2018 Contentful, this is a BETA release'
+      'Copyright 2019 Contentful, this is a BETA release'
     ].join('\n'))
 }
 
@@ -46,7 +46,7 @@ export async function environmentAliasList ({ context }) {
   const aliases = result.items.sort((a, b) => a.name.localeCompare(b.name))
 
   const table = new Table({
-    head: ['Alias id', 'Aliased to']
+    head: ['Alias id', 'Target']
   })
   aliases.forEach((alias) => {
     table.push([
--- a/lib/cmds/space_cmds/alias_cmds/list.js
+++ b/lib/cmds/space_cmds/alias_cmds/list.js
@@ -1,0 +1,62 @@
+
+import Table from 'cli-table3'
+
+import { handleAsyncError as handle } from '../../../utils/async'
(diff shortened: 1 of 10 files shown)


lib/utils/pagination.js
```diff
@@ -12,7 +12,7 @@ export default async function paginate ({client, method, skip = 0, limit = 100,
     },
     query
   )
-
+debugger
```
jroehl: this is prob. by mistake, right?

lib/cmds/space_cmds/alias_cmds/list.js
```diff
@@ -0,0 +1,61 @@
...
+      type: 'string'
+    })
+    .option('management-token', {
+      alias: 'mt',
+      describe: 'Contentful management API token',
+      type: 'string'
+    })
+    .epilog([
+      'Environment aliases is an [ALPHA] Feature',
+      'Copyright 2018 Contentful, this is a BETA release'
```
phoebeschmidt: copyright 2019? this is an alpha release?

lib/cmds/space_cmds/alias_cmds/list.js
```diff
@@ -0,0 +1,61 @@
...
+  })
+
+  const space = await client.getSpace(activeSpaceId)
+
+  const result = await paginate({ client: space, method: 'getEnvironmentAliases' })
+
+  const aliases = result.items.sort((a, b) => a.name.localeCompare(b.name))
+
+  const table = new Table({
+    head: ['Alias id', 'Aliased to']
```
phoebeschmidt: might need to change `Aliased to` to `Targets`, regarding our convo about terminology with marketing folks yesterday

lib/cmds/space_cmds/alias_cmds/list.js
```diff
@@ -0,0 +1,61 @@
...
+    })
+    .epilog([
+      'Environment aliases is an [ALPHA] Feature',
+      'Copyright 2018 Contentful, this is a BETA release'
+    ].join('\n'))
+}
+
+export const aliases = ['ls']
+
+export async function environmentAliasList ({ context }) {
```
phoebeschmidt: do we need to also destructure the spaceId here if it's passed in by the user? 

(my understanding is that `context` is anything saved in the config, but not any options passed in when the user uses the command?)

lib/cmds/space_cmds/alias_cmds/update.js
```diff
@@ -0,0 +1,58 @@
...
+    })
+    .epilog([
+      'See more at:',
+      'Environment ALiases is an [ALPHA] feature',
+      'https://github.com/contentful/contentful-cli/tree/master/docs/space/alias/update',
+      'Copyright 2018 Contentful, this is a BETA release'
+    ].join('\n'))
+}
+
+export async function environmentAliasUpdate ({ context, aliasId, environmentId }) {
```
phoebeschmidt: do we also need to destructure the spaceId here if it's passed in by the user? (same q as above)

test/unit/cmds/space_cmds/alias_cmds/update.test.js
```diff
@@ -0,0 +1,58 @@
...
+  managementToken: 'mockedToken'
+})
+
+afterEach(() => {
+  createManagementClient.mockClear()
+  getEnvironmentAliasStub.mockClear()
+  log.mockClear()
+})
+
+test.only('list environment aliases', async () => {
```
phoebeschmidt: remove the `.only` and update the name of the test pls :)

lib/utils/pagination.js
Khaledgarbaya: It's an easter egg 🕺

lib/cmds/space_cmds/alias_cmds/list.js
Khaledgarbaya: Maybe `Target` instead?

lib/cmds/space_cmds/alias_cmds/list.js
Khaledgarbaya: Thanks to middleware activeSpaceId will always haver the correct space ID whether it's passed or set in the config

lib/cmds/space_cmds/alias_cmds/update.js
Khaledgarbaya: the middleware will take care of that and make the space available through `context.activeSpaceId`

lib/cmds/space_cmds/alias_cmds/update.js
```diff
@@ -0,0 +1,58 @@
...
+
+export const builder = (yargs) => {
+  return yargs
+    .usage('Usage: contentful space alias update --alias-id master --environment-id staging')
+    .option('alias-id', {
+      alias: 'a',
+      describe: 'Id of the alias to create',
+      demandOption: true
+    })
+    .option('environment-id', {
```
phoebeschmidt: I wonder if we should call this something else. `target-environment-id`? just because we use --environment-id everywhere else to mean the environment we are acting on?

IDK i'm torn about this

lib/cmds/space_cmds/alias_cmds/update.js
phoebeschmidt: hm then the middleware also takes care of `environmentId` right?

lib/cmds/space_cmds/alias_cmds/update.js
Khaledgarbaya: That's a valid point

lib/cmds/space_cmds/alias_cmds/update.js
```diff
@@ -0,0 +1,58 @@
...
+import { handleAsyncError as handle } from '../../../utils/async'
+import { createManagementClient } from '../../../utils/contentful-clients'
+
+export const command = 'update'
+
+export const desc = 'Update an environment alias'
+
+export const builder = (yargs) => {
+  return yargs
+    .usage('Usage: contentful space alias update --alias-id master --environment-id staging')
```
phoebeschmidt: let's just change the `usage` to include `--target-environment-id` :)

PR #178, opened 2019-09-27
chore(cleanup): Add prettier and remove babel

jroehl:
## Summary

Remove transpilation and introduce prettier.

## Description

- removed babel and transpilation setup
- changed custom eslint configuration to prettier

## Motivation and Context

To simplify keeping dependencies up to date and remove unnecessary complexity from the build/lint/develop process, we can remove babel, a custom eslint config from the codebase. Instead we can rely on an opinionated linter (`prettier`) and use the built-in features of node >= 8, getting rid of the transpilation process.

## Todos

- [x] Implemented feature


.prettierrc
```diff
@@ -0,0 +1,3 @@
+{
+  "singleQuote": true
```
Khaledgarbaya: our codebase for the other SDKs and tools are not using `;` I don't have a strong opinion about it but maybe we should keep the same behaviour

jroehl: Yepp - that's true, it's just a prettier setting .. I don't mind changing it

jroehl: (I am just used to semicolons 🤓)

lib/utils/github.js
```diff
@@ -1,63 +1,62 @@
...
   return new Listr([
     {
       title: `Fetching release information of ${repo}`,
-      task: async (ctx) => {
-        try {
-          const response = await axios({
-            url: `https://api.github.com/repos/${repo}/releases/latest`
-          })
-          ctx.latestReleaseInfo = response.data
-        } catch (err) {
```
Khaledgarbaya: 🤦‍♂

lib/utils/github.js
```diff
@@ -1,63 +1,62 @@
...
-      task: async (ctx) => {
-        ctx.latestReleaseZipLocation = path.join(destination, 'latest-release.tar.gz')
-        try {
-          const response = await axios({
-            url: ctx.latestReleaseInfo.tarball_url,
-            responseType: 'stream'
-          })
-          ctx.latestReleaseTarballStream = response.data
-        } catch (err) {
-          throw err
```
Khaledgarbaya: 😅

jroehl: I love it 😬

lib/cli.js
```diff
@@ -1,15 +1,16 @@
-import yargs from 'yargs'
-import {log} from './utils/log'
-import {buildContext, getCommand, assertContext} from './utils/middlewares'
-import {version} from '../package.json'
+const yargs = require('yargs')
```
makinwab: Any reason we're not using the es6 import and export style?

Khaledgarbaya:
@makinwab 
> Have one question which I've written below and a suggestion to use chore instead of feat when squashing commits and merging the PR.

If we use chore we won't have a new release with the changes. However, one can argue that this is not a suer facing release and it can wait to another proper release

lib/cli.js
jroehl: By using commonjs we can get rid of the babel transpilation step.. It's currently the only reason we have to transpile the code, all the other features are included from Node >= 8 anyways.

makinwab:
> @makinwab
> 
> > Have one question which I've written below and a suggestion to use chore instead of feat when squashing commits and merging the PR.
> 
> If we use chore we won't have a new release with the changes. However, one can argue that this is not a suer facing release and it can wait to another proper release

Yeah, that's my point @Khaledgarbaya. Not to have a release based on changes that don't affect the users and it's really a chore.

lib/cli.js
makinwab: Got it. 😢

PR #187, opened 2019-10-23
fix(typo): correct grammer of json space import cli option

joelhoelting:
## Description

For the command `contentful space import` options, `--content-file` has a grammatically incorrect description. 

"JSON file that contains data to be import to your space" is incorrect English. It should be something along the lines of "JSON file that contains data to import into a space" or "JSON file that contains data to be imported into a space". The former is a bit cleaner.

## Motivation and Context

It correctly a typo in the CLI options


jroehl:
Thanks for your PR, will merge it as `docs()`, so that it will be included in the next release. 🙇

PR #188, opened 2019-10-29
update migration doc

llwor94:
## Summary

updated migration doc to display correct management token command, as -a and --access-token no longer work


makinwab:
Thanks for helping with this. 👍🏾

PR #207, opened 2020-01-08
fix(config): Require non-empty string values for config options

ruderngespra:
Require non-empty string values for config options so that the cli no longer accepts commands like `contentful config add --mt --as --ae`


lib/cmds/config_cmds/add.js
```diff
@@ -42,7 +42,16 @@ module.exports.builder = yargs => {
         'Pass proxy config as raw config instead of creating a httpsAgent',
       type: 'boolean'
     })
+    .requiresArg([
+      'management-token',
+      'active-space-id',
+      'active-environment-id'
+    ])
     .config('config', 'set all the config options at once')
+    .fail(msg => {
```
jroehl: We should consider using the logging utility from `lib/utils/log.js` -> `logError` instead of calling `console.log` directly.

PR #305, opened 2020-04-03
fix(mkdirp): adopt latest version change

makinwab:


makinwab:
> I expect you'll also have to change it in:
> `lib/utils/github.js` and `test/unit/utils/fs.test.js`

github.js has been updated, yes, however, we still need to promisify `fs.mkdir` based on [the doc](https://nodejs.org/api/fs.html#fs_fs_mkdir_path_options_callback)

PR #463, opened 2020-06-19
docs(help): Fix management token typo

floftar:
## Summary

Management token option can be given as "management-Token",
but the project uses "management-token" everywhere else.

## Motivation and Context

For consistency it is better to have same instructions everywhere.


makinwab:
Thanks for helping with the typo, @floftar.

PR #502, opened 2020-07-13
fix(login-input): Mask token paste input

kdamball:
## Summary

Masks token pasting when logging in

## Description

Uses input type "password" instead of "input" as we're dealing with secret tokens.

## Motivation and Context

Security focused where tokens aren't availabe in CLI output logs.


kdamball:
@jroehl Do I need to manually bump the version?

jroehl:
@kdamball If you're merging this with a semver commit message e.g `fix(input): Mask ..` it should be released automatically..

PR #555, opened 2020-08-18
fix(args): proxy must be passed as a string to export/import/migration

matthew-contentful:
Proxy args are represented internally as an object in most cases (when loaded from config, when loaded from the ENV..), but must be passed to other cli tools (export/import/migration) as strings. In the case where the proxy arg is from the command line and the call is for export/import (migration doesn't support the proxy command line arg) the string arg was treated as an object, and so using the arg would lead to an error. This change passes the string directly.

(the unrelated change with management* is just to make export more closely match import)

### DIFF
--- a/lib/cmds/space_cmds/export.js
+++ b/lib/cmds/space_cmds/export.js
@@ -142,15 +142,13 @@ const exportSpace = async argv => {
     proxy,
     rawProxy
   } = context
-  const managementApplication = `contentful.cli/${version}`
-  const managementFeature = feature
 
   const options = {
     ...argv,
     spaceId: activeSpaceId,
     environmentId: activeEnvironmentId,
-    managementApplication,
-    managementFeature,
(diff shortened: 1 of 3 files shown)


lib/cmds/space_cmds/export.js
```diff
@@ -159,9 +157,15 @@ const exportSpace = async argv => {
     // contentful-import and contentful-export
     // expect a string for the proxy config
     // and create agents from it
-    options.proxy = proxyObjectToString(proxy)
+    if (typeof proxy !== 'string') {
```
phoebeschmidt, line 160: why not put this check in the `proxyObjectToString` function itself? :)

matthew-contentful: I updated the proxyObjectToString to confirm that the arg is correct, but the check here seems correct to me because it distinguishes between two correct cases.

phoebeschmidt (approved):

PR #605, opened 2020-09-23
fix(config): add no output timeout to bump-brew-formula CI step

phoebeschmidt:
Finally fix this brew formula thing since upgrading the xcode version there.


marcolink:
lgtm 🎉

PR #723, opened 2020-12-11
fix(ci): address shallow cloning for homebrew

makinwab:
Homebrew-core is now a shallow clone in our brew step, now we explicitly fetch with the --unshallow option 
Context: https://github.com/Homebrew/brew/pull/9383


matthew-contentful (approved):
LGTM

PR #783, opened 2021-02-08
fix(export-assets): bump contentful export version

Vinz93:
## Summary

Bump contentful export version:

This export version will handle assets that were not successfully processed and notify the user of which of them had issues.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3324,9 +3324,9 @@
       }
     },
     "contentful-export": {
-      "version": "7.11.4",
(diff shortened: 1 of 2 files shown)


ruderngespra (approved):

ruderngespra (approved):

ruderngespra (changes requested):
The `package-lock.json` file is expanded a lot, I think because you use npm version 7. Let's align on if we want that to happen or not.

ruderngespra (approved):

PR #868, opened 2021-04-07
feat: support passing additional HTTP headers

z0al:
## Summary

Support passing additional HTTP Headers to (most) commands. For example:

```sh
contentful export -H "x-contentful-enable-alpha-feature: my-feature" ...args
```

## Description

Added `--header/-H` option to the following command namespaces:

- Content Type
- Extension
- Organization
- Space

A header value should be `"key: value"` format. Whitespaces are allowed but will be trimmed. A value that doesn't follow the aforementioned format will be ignored silently. 

## Motivation and Context

Enabling users to pass additional headers can be useful in general but especially when testing out alpha features (e.g. part of EAP program)

## Todos

- [x] Support `--header/-H` option in necessary commands
- [ ] Update docs. Is this supposed to be done manually?
- [x] Fix & add more tests if necessary

P.S. Sorry about the changes set. Most of it caused by running `npm run lint -- --fix`. I had to adjust Prettier rules that are mostly ignored throughout the code to get a smaller changeset (was 133 files before 🤯 )

### DIFF
--- a/lib/cmds/space_cmds/generate_cmds/migration.js
+++ b/lib/cmds/space_cmds/generate_cmds/migration.js
@@ -9,6 +9,7 @@ const {
   createManagementClient: createClient
 } = require('../../../utils/contentful-clients')
 const fs = require('fs')
+const { getHeadersFromOption } = require('../../../utils/headers')
 
 const b = recast.types.builders
 
@@ -47,6 +48,11 @@ module.exports.builder = yargs => {
       alias: 'f',
       type: 'string'
     })
+    .option('header', {
+      alias: 'H',
+      type: 'string',
+      describe: 'Pass additional HTTP Header'
+    })
     .epilog(
       [
         'See more at:',
@@ -58,7 +64,7 @@ module.exports.builder = yargs => {
 
 module.exports.aliases = ['m']
 
-const ctNameNeedsEscaping = function(ctId) {
+const ctNameNeedsEscaping = function (ctId) {
   const reservedWords = [
     'break',
     'case',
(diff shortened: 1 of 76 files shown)


z0al:
The `/docs` folder seems out of sync to me, what is the status there? Also, do  I have to update the it manually?

lib/cmds/space_cmds/generate_cmds/migration.js
```diff
@@ -323,21 +329,27 @@ const generateMigrationScript = async function(environment, contentTypes) {
...
 }
 
-const getEnvironment = async function(managementToken, spaceId, environmentId) {
-  const client = await createManagementClient(managementToken)
+const getEnvironment = async function (
+  managementToken,
+  spaceId,
+  environmentId,
+  headers
+) {
```
damienxy, line 351: I would turn these params into an object now that the argument list is getting long

marcolink:
@z0al great work!
I was just wondering if the arg should be plural e.g. `headers`? The way I understand it, you can also pass a list of headers, right?

z0al:
> @z0al great work!
> I was just wondering if the arg should be plural e.g. `headers`? The way I understand it, you can also pass a list of headers, right?

@marcolink you can pass only one value per option though. To pass multiple values you need to repeat the option (`-H/--header`) e.g.

```sh
cli -H "header1: value" -H "header2: value" ...
```

I followed `curl` example here and don't really have a strong opinion on the matter.

marcolink:
> > @z0al great work!
> > I was just wondering if the arg should be plural e.g. `headers`? The way I understand it, you can also pass a list of headers, right?
> 
> @marcolink you can pass only one value per option though. To pass multiple values you need to repeat the option (`-H/--header`) e.g.
> 
> ```shell
> cli -H "header1: value" -H "header2: value" ...
> ```
> 
> I followed `curl` example here and don't really have a strong opinion on the matter.

Ah, I thought one would pass in a comma-separated list of headers 💡.

Then the current naming is totally fine :)

z0al:
> Ah, I thought one would pass in a comma-separated list of headers bulb.

That will require a bit more sophisticated header parsing than a simple string split :)

marcolink (approved):
:shipit:

PR #888, opened 2021-04-19
chore: release experimental branch

anho:
## Summary

In order to properly use other experimental versions of our packages, we need a new channel here too.


marcolink (approved):
:shipit:

PR #1005, opened 2021-06-14
feat(env-creation): increase retry default for env processing check

ruderngespra:
To account for environments for which creation takes longer than 30 seconds, we increase the number of retries that environment processing status is checked.


matthew-contentful (approved):
LGTM

Issue #831, opened 2021-03-12
--await-processing option not reliable when creating an environment


damienlepage:
## Expected Behavior

We have a deployment script that clones the master environment into a test environment before running a migration:

```
contentful space environment create --space-id $space --environment-id $staging_env --name $staging_env --source 'master' --await-processing true
contentful space migration --space-id $space --environment-id $staging_env --yes $migration_file
```

The expectation is that the `--await-processing true` option on the first command will allow the migration on the second command to work. 

## Actual Behavior

The migration step very often fails with the following error:

```
│ Error: The provided space does not exist or you do not have access.                                                                                                                                                                                                          │
│                                                                                                                                                                                                                                                                              │
│ 🚨  Migration unsuccessful                                                                                                                                                                                                                                                   │
│     at run (/snapshot/project/node_modules/contentful-migration/built/bin/cli.js:121:23)                                                                                                                                                                                     │
│     at processTicksAndRejections (internal/process/task_queues.js:97:5)                                                                                                                                                                                                      │
└──────────────────────────────────────────────────────────────────────────
```

We tried adding a `sleep 10` between the 2 steps and it is still often not enough.

## Environment

Contentful-cli: 1.5.35
Node Version: v15.5.0
Npm Version: 7.3.0

Vinz93:
Hi @damienlepage 👋  We added a new [flag](https://github.com/contentful/contentful-cli/pull/1006) `--processing-timeout`, now you can specify how long the awaiting processing should last, you can find it our last version contentful-cli: 1.8.0

PR #1006, opened 2021-06-14
feat(env-creation): add processing-timeout option

ruderngespra:
Add optional option to provide a specified timeout in minutes for `--await-processing` upon environment creation.

**Open Issues**
- We should add a test [here](https://github.com/contentful/contentful-cli/blob/master/test/unit/cmds/space_cmds/environment_cmds/create.test.js) that checks that `environmentCreate` triggers `checkAwaitProcessing` with the correct parameters. How to mock that correctly?
- Snapshot tests behave flaky locally.

### DIFF
--- a/lib/cmds/space_cmds/environment_cmds/create.js
+++ b/lib/cmds/space_cmds/environment_cmds/create.js
@@ -56,7 +56,6 @@ module.exports.builder = yargs => {
       type: 'string',
       describe: 'Pass an additional HTTP Header'
     })
-    .implies('processing-timeout', 'await-processing')
     .epilog(
       [
         'See more at:',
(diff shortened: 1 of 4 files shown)


Vinz93 (approved):
LGTM :)

lib/cmds/space_cmds/environment_cmds/create.js
```diff
@@ -88,35 +124,12 @@ async function environmentCreate({
...
-        } else {
-          logging.error('Environment creation failed')
-        }
-        break
-      }
-
-      await Promise.delay(DELAY)
-      count++
-    }
+    checkAwaitProcessing(space, environment, processingTimeout)
```
makinwab, line 132: We should log some type of timeout-related message for users after awaiting processing in the case where the environment is not ready. This way the user knows this and can bump their timeout values.

Vinz93: Thank you for the feedback, I'll this change in another PR 👍

Nintendot:
🎉 💥 !

Nintendot:
Would this PR open doors for addressing https://github.com/contentful/contentful-cli/issues/831?

Vinz93:
> Would this PR close #831?

Yes! 🙂

PR #1028, opened 2021-06-22
fix(env-creation): awaiting processing timeout message

Vinz93:
Informe users when environment processing is not ready and the awaiting time is over


ruderngespra (approved):
lgtm :)

PR #1266, opened 2021-12-07
fix(tests): Fix broken tests

andreascful:
Fix tests that were broken by faulty Dependabot merges (e.g. this [PR](https://github.com/contentful/contentful-cli/pull/1127)).


marcolink:
@andreascful should we add those to the ignore list? https://github.com/contentful/contentful-cli/blob/master/.github/dependabot.yml#L11

PR #1274, opened 2021-12-08
chore(deps): bump internal dependencies to support standard header

andreascful:
Bump contentful-import, contentful-export and contentful-migration versions to current to support standard header.


marcolink (approved):
:shipit:

PR #1302, opened 2022-01-04
feat(cmds): Remove boilerplate command [SDK-778]

andreascful:
## Summary

Removes all instances and tests of the boilerplate command because the related API endpoint will be [deprecated on the 19th of January](https://www.contentful.com/developers/api-changes/#deprecating-boilerpate-api-endpoint-and-cli-command).


docs/guide/README.md
```diff
@@ -2,8 +2,7 @@
 
 A guide introducing basic concepts for working with Contentful.
 
-This command is intended for users new to Contentful, regular users can run
-`contentful boilerplate` directly to quickly set up an example project.
+This command is intended for users new to Contentful.
```
marcolink, line 5: why would we keep this line?

marcolink (approved):

docs/guide/README.md
```diff
@@ -23,7 +22,7 @@ and add the first Content entries to the freshly created space.
 
 #### 4. Download app to display your content
 
-A shortcut to [contentful boilerplate](../boilerplate) to download a boilerplate
+A shortcut to download a [boilerplate example project](https://github.com/contentful?q=boilerplate&type=all&language=&sort=) to download a boilerplate
 example project built with one of our SDKs.
```
damienxy, line 26: Is this still correct? 
It mentions boilerplate, but links to Contentful GitHub (and there's some duplicate text in the new version)

damienxy (approved):
Do we have a changelog somewhere? Would be good to add it there

docs/guide/README.md
andreascful: I wasn't sure about it either but now when reading it a second time I think we should remove this line completely. Thank you

docs/guide/README.md
andreascful: Good question. I changed the text now to be more meaningful. Would you leave it like it is in the new version or remove it altogether?

damienxy: Sounds great now 👍

Issue #1332, opened 2022-01-26
Experimental version of CLI doesn't support `defaultValue` from migration DSL


dustinbrownman:
## Expected Behavior

We recently started using Compose to manage some of our content creation. We are trying to create a migration to add a new page type, following this helpful guide: https://www.contentful.com/developers/docs/compose/cli-tools/.

One of the content models we are adding as part of this migration has a default value. I expected to not get any validation errors when running the migration.

## Actual Behavior

When I run the migration, I get a validation error saying that `defaultValue` is not a valid property name for a field. Here's the error:

```
Validation failed


Errors in /path/to/migration/script

Line 182: "defaultValue" is not a valid property name for a field.
180:     .type("Symbol")
181:     .required(true)
182:     .defaultValue({
183:       "en-US": "Some default value",
184:     });

🚨  Migration unsuccessful
🚨  Error: Step Validation Errors
```

## Possible Solution

In order to add `{ relationshipType: ['Composition'] }` to the validations which creates the link in Compose, we need to use the experimental version of the CLI, specifically [v1.6.3-experimental.2](https://github.com/contentful/contentful-cli/tree/v1.6.3-experimental.2).

That version of the CLI uses [v4.2.0-experimental.1](https://github.com/contentful/contentful-migration/blob/v4.2.0-experimental.1/README.md) of the contentful-migration library. Looking through that version of the library, it doesn't appear to have support for `defaultValue`.

The experimental version of the CLI appears to be about 6 months old. The suggested solution is to cut a new release that supports the more updated version of the migration DSL.

## Steps to Reproduce

1. Install (and make sure you're using) the experimental version of the CLI
2. Add a migration that creates a field on some content type
```javascript
const someContentType = migration.editContentType('someContentType')

someContentType.createField("something")
    .name("Something")
    .type("Symbol")
    .defaultValue({
        "en-US": "Something goes here",
    });
```
3. Run the migration and notice the error

## Context

Hopefully the there's enough context above, but let me know if I can provide more 😄 

## Environment

-   **Language Version**: v17.3.1
-   **Package Manager Version**: 8.3.0
-   **Operating System**: Darwin Kernel Version 21.2.0: Sun Nov 28 20:29:10 PST 2021; root:xnu-8019.61.5~1/RELEASE_ARM64_T8101 arm64
-   **Package Version**: contentful-cli@1.6.3-experimental.2

dustinbrownman:
[...]

PR #1347, opened 2022-02-04
Bump contentful-migration version

dustinbrownman:
## Summary

This bumps the experimental version of https://github.com/contentful/contentful-migration being used on the experimental branch.

**Note:** The `lockfileVersion` was updated as part of this, which is why that diff is so big. It looks like the `master` branch also uses the newer lockfile version, so it seems like a good idea to update it on this branch as well, but let me know if you'd like me to use an older version of NPM so the diff is simpler.

## Motivation and Context

The [latest experimental version](https://github.com/contentful/contentful-migration/releases/tag/v4.2.0-experimental.5) is more up-to-date with the current migration DSL. Notably, `defaultValue` is supported in this version and isn't in the [current version](https://github.com/contentful/contentful-migration/releases/tag/v4.2.0-experimental.1). This limits users who are wanting to take advantage of some of the experimental features (e.g. working with Compose). Here are a couple of documented issues for example:

- https://github.com/contentful/contentful-cli/issues/1332
- https://github.com/contentful/contentful-migration/issues/998


ywenc (approved):

dustinbrownman:
@damienxy @andreascful I'm not sure if you're the correct folks to ping about this, but it looks like the two of you are the only humans to have recent commits to this project. I was hoping we could get some eyes on this. We want to invest more of our resources into Contentful, but hiccups like the experimental version being out of date is giving us trouble. Please let me know if anything needs to change!

andreascful:
Hello Dustin,

thank you for your suggested change. The change looks good to me and we can merge it into the experimental branch. I only changed one of the tests so that the build won't break on us (due to deprecation of the boilerplate api).
However, please be aware that the experimental branch might change in the future and we don't recommend using it in production.

Best wishes
Andreas

dustinbrownman:
@andreascful Thank you! This will help us a ton as we migrate more content over to Contentful.

> However, please be aware that the experimental branch might change in the future and we don't recommend using it in production.

This totally makes sense. Do you know if there's a timeline for when the Compose features will be available in the main releases of the CLI and contentul-migration? We've found Compose to be a handy tool. Losing those relationships because they are incompatible with the current tooling would be a bummer.

PR #1397, opened 2022-03-14
chore(test): Fix tests

andreascful:
Fix unit tests. Due to a missconfigured command most of the unit tests were not being executed.


marcolink (approved):
Thank you @andreascful 🏅

PR #1399, opened 2022-03-15
feat(params): Add delivery-token param to export cmd [ZEND-1938]

andreascful:
Add delivery-token param to the param list for the export command. We already support this command but it was not listed so far.


marcolink (approved):
looking good 🚢

PR #1423, opened 2022-04-05
chore(docs): add retry-limit and timeout parameter to docs []

ruderngespra:
Update docs to account for missing parameters.

- double check that `.contentfulrc` values won't be overwritten by newly introduced default. If so, revert. (Reverted.)

### DIFF
--- a/docs/space/import/README.md
+++ b/docs/space/import/README.md
@@ -10,11 +10,10 @@ Usage: contentful space import --content-file <file>
 Options:
   -h, --help                 Show help                                 [boolean]
   --space-id                 ID of the destination space                [string]
-  --environment-id           ID the environment in the destination space
-                                                    [string] [default: "master"]
-  --management-token         Contentful management API token            [string]
-  --content-file             JSON file that contains data to be import to your
-                             space                           [string] [required]
+  --environment-id           ID the environment in the destination space[string]
+  --management-token, --mt   Contentful management API token            [string]
+  --content-file             JSON file that contains data to import into a space
+                                                             [string] [required]
   --content-model-only       Import only content types[boolean] [default: false]
   --skip-content-model       Skip importing content types and locales
                                                       [boolean] [default: false]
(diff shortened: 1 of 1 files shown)


docs/space/import/README.md
```diff
@@ -10,23 +10,27 @@ Usage: contentful space import --content-file <file>
 Options:
   -h, --help                 Show help                                 [boolean]
   --space-id                 ID of the destination space                [string]
-  --environment-id           ID the environment in the destination space
-                                                    [string] [default: "master"]
-  --management-token         Contentful management API token            [string]
-  --content-file             JSON file that contains data to be import to your
-                             space                           [string] [required]
+  --environment-id           ID the environment in the destination space[string]
```
marcolink, line 13: did we remove the default value here?

docs/space/import/README.md
```diff
@@ -10,23 +10,27 @@ Usage: contentful space import --content-file <file>
...
+                                                             [string] [required]
   --content-model-only       Import only content types[boolean] [default: false]
   --skip-content-model       Skip importing content types and locales
                                                       [boolean] [default: false]
   --skip-locales             Skip importing locales   [boolean] [default: false]
   --skip-content-publishing  Skips content publishing. Creates content but does
                              not publish it           [boolean] [default: false]
   --error-log-file           Full path to the error log file            [string]
-  --management-host          Management API host
+  --host                     Management API host
```
marcolink, line 24: is this a breaking change?

docs/space/import/README.md
ruderngespra: I think this is a leftover from the transition from the import cli to the generic cli where the default never was added (see https://github.com/contentful/contentful-import/blob/master/lib/usageParams.js#L12 and https://github.com/contentful/contentful-cli/blob/master/lib/cmds/space_cmds/import.js. But good point, the environment id option is out of sync between contentful-import and contentful-cli which is really a bit of a pain :) Not sure if the default from contentful-import gets picked up 🤔. I'll test it and try to align.

ruderngespra: Looks to me as if even though we don't explicitly set the default here anymore, we still default to `master` when setting the context here: https://github.com/contentful/contentful-cli/blob/master/lib/utils/middlewares.js#L42

We "removed" the default value when we moved everthing from the import-cli to the contentful-cli, but in practice, we'll still default to master if I see this correctly. In order to stay consistent, I added the default again to the cli option.

marcolink (approved):

docs/space/import/README.md
ruderngespra: Had to revert this change, as this would actually have introduced a breaking change I overlooked, as now we always would have overwritten `activeEnvironmentId` from the `.contentfulrc` config with `master`. So actually `master` has never been the default for the import command since it moved to contentful-cli and `.contentfulrc` was introduced, so the updated docs are correct in not saying so.

marcolink: does it mean if you don't provide a `.contenfulrc` file with an env id, the tool fails to execute? 
WHich makes sense to me, as you're not accidentally importing into your production env 🤔

marcolink (approved):

PR #1486, opened 2022-05-18
fix: copyright content 

mshaaban0:
## Summary

Contentful cli outputs the wrong year `2019`. 
This change prints out the correct current year and centralise the content

**Before**
Screenshot 2022-05-18 at 17 18 28

**After**
Screenshot 2022-05-18 at 17 17 58


ruderngespra (changes requested):
Thanks for tackling this! :) I just wonder if there is a way to make sure the snapshots won't create failing tests once we enter 2023 🤔.

marcolink:
hey @mshaaban0 should we bring this over the finish line? :)

denkristoffer:
We can remove the year from the statement, it shouldn't be a required part of a copyright statement. But if we go with that, let's ask legal first.

marcolink:
@mshaaban0 should we get it over the finish line? :)

test/e2e/__snapshots__/basics.test.js.snap
```diff
@@ -17,7 +17,7 @@ Options:
   -h, --help     Show help                                             [boolean]
   -v, --version  Show current version                                  [boolean]
 
-Copyright 2019 Contentful"
+Copyright 2013 Contentful"
```
marcolink, line 20: Not 100% sure about 2013 here 🤔

mshaaban0: When Contentful was founded ;)

mshaaban0: Only for snapshot testing though, the current year will be printed for users

marcolink (approved):

ruderngespra (approved):
👍🏻 🎉

PR #1518, opened 2022-06-07
chore(deps): update contentful-import and contentful-export

ToppleTheNun:
Updates contentful-import and contentful-export to latest to fix date-fns issue


marcolink (approved):
Thanks @ToppleTheNun for going through the full dependency tree with all the updates 🎉

ToppleTheNun:
Thanks @Brad-Turner for the original work.

PR #1556, opened 2022-07-19
refactor: remove guide command

mshaaban0:
## Summary

Remove guide command in favour of the upcoming `init` command

## Motivation and Context

We will introduce a new `init` command that will onboard users to Contentful. Hence we don't need the `guide` anymore as it's flow is broken

### DIFF
--- a/test/integration/cmds/logout.test.js
+++ b/test/integration/cmds/logout.test.js
@@ -12,7 +12,7 @@ beforeAll(() => {
   return initConfig()
 })
 
-test.only('should not logout', done => {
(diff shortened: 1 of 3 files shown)


test/integration/cmds/logout.test.js
```diff
@@ -12,7 +12,7 @@ beforeAll(() => {
   return initConfig()
 })
 
-test('should not logout', done => {
+test.only('should not logout', done => {
```
denkristoffer, line 15: Is this intended?

PR #1565, opened 2022-07-26
feat: spaces list improvements [BAU-856]

bgutsol:
- spaces list command improvements
- add activeOrganizationId to the global context


mshaaban0 (commented):
Thanks for doing the changes. There are a couple of things we need to address before this goes in:

**Active Space and Org names:** Should always be on the top of the list "Colored"
**Filtration:** I couldn't grep the output of `ls` option `contentful space ls | grep SPACE_NAME_SEARCH_TEXT`
**Pagination:**
* The output overrides everything on the screen when used more than once
* User should be able to navigate `next`, and `prev` preferably with `n` and `p` keys

mshaaban0:
Screenshot 2022-07-26 at 17 38 02
Screenshot 2022-07-27 at 09 58 24

PR #1583, opened 2022-08-11
feat: add create new space and populate content [BAU-890]

massao:
## Summary

* Migrate create space and import space to typescript
* Removed default locale as it is missing on type definition on contentful-import
* Added functionality on init to create a new space and populate with content.
* Added content file to be used when populating new space.

## How to verify locally

**New space with content in init:**
run npm link && npm tsc:watch
then run contentful init
You should be asked to continue login to Contentful (if you're not logged in)
After login, you will be asked to create a new space or use existing one(currently not working)
And after selecting to create a new one, it will ask if you want to have content added to the new space
Ask for user confirmation before creating space, and in which organization to create if no organization is is present.

**Jest and Typescript**
Set [env variables](https://github.com/contentful/contentful-cli#robot-testing) locally
then run `test/unit/cmds/space_cmds/create.test.ts test/unit/cmds/space_cmds/import.test.ts`
Tests should run and pass ✅


mshaaban0:
Hey @massao It seems like we have some ts compilation errors when running `npm run tsc:watch` on this branch

Screenshot 2022-08-18 at 14 29 05

mshaaban0 (approved):
Silky Smooth 💯 ❤️ Thank you. 

Screenshot 2022-08-19 at 16 27 47

lib/cmds/space_cmds/create.ts
```diff
@@ -103,11 +106,7 @@ the Pricing page: https://www.contentful.com/pricing/?faq_category=payments&faq=
   }
 
   const intentSystem = new IntentSystem()
-  intentSystem.addHandler(
-    createSpaceIntents({
-      skipConfirm: yes
-    })
-  )
+  intentSystem.addHandler(createSpaceIntents())
```
mshaaban0, line 109: Why did we remove `skipConfirm: yes` here?

massao: my bad, forgot to comment about this one, it is because the `createSpaceIntents` doesn't accept any parameters, I saw that the previous version of the method accepted, but it was removed from there, but not here:
https://github.com/contentful/contentful-cli/blob/master/lib/core/event-handlers/intents/create-space-handler.js

lib/cmds/space_cmds/create.ts
```diff
@@ -43,11 +44,6 @@ module.exports.builder = yargs => {
       describe:
         'Confirm space creation without prompt, be aware this may result in extra monthly charges depend on your subscription'
     })
-    .option('default-locale', {
```
mshaaban0, line 46: @massao this seems to be removed by mistake?

massao: Actually this is because the `createSpace` function doesn't accept the `defaultLocale` param.
As you can see here: https://github.com/contentful/contentful-management.js/blob/ea3b20fd431544b9a35fc40ee23c3f7c747d3383/lib/create-contentful-api.ts#L102-L105
It expects `spaceData: Omit<SpaceProps, 'sys'>`
And from here you can see that `SpaceProps` having the `sys` omitted is just `name`: https://github.com/contentful/contentful-management.js/blob/ea3b20fd431544b9a35fc40ee23c3f7c747d3383/lib/entities/space.ts#L8-L11

mshaaban0: huh, good catch. so weird it was never working! cc @marcolink

lib/cmds/init.ts
```diff
@@ -15,14 +19,66 @@ export const builder = (yargs: Argv) => {
 }
 
 export const init = async () => {
+  greetings()
+
   const context = await getContext()
+  const managementToken = context.managementToken
 
-  greetings()
+  const token = await login({
```
mshaaban0, line 27: we should call login here only if `(!context.managementToken)` doesn't exist

mshaaban0 (approved):
Looks awesome 💯

mshaaban0 (approved):

lib/cmds/space_cmds/create.ts
mshaaban0: True, it was being [passed](https://github.com/contentful/contentful-management.js/blob/master/lib/create-contentful-api.ts#L110) nonetheless. We will to put it back in another PR, thanks.

Here's a PR to fix the types https://github.com/contentful/contentful-management.js/pull/1474

PR #1595, opened 2022-08-17
feat: add select space method to init

massao:
## Summary

* Migrate use space to typescript
* Added functionality on init if use decide to use existing space.

## How to verify locally

**User select existing space**
run npm link && npm tsc:watch
then run contentful init
You should be asked to continue login to Contentful (if you're not logged in)
After login, you will be asked to create a new space or use existing one
Select use existing one, and a list with the spaces you can choose will appear

**Jest and Typescript**
Set [env variables](https://github.com/contentful/contentful-cli#robot-testing) locally
then run `npx jest test/unit/cmds/space_cmds/use.test.ts`
Tests should run and pass ✅


lib/cmds/space_cmds/use.ts
```diff
@@ -31,7 +32,14 @@ module.exports.builder = yargs => {
     .epilog('Copyright 2019 Contentful')
 }
 
-function showSuccess(space, env) {
+interface Space {
```
denkristoffer, line 35: Can we use `@contentful/types` maybe?

massao: I'll take a look into it.

massao: Had to import from `contentful-management`, because it seems like the types are different, from this package and the `@contentful/types` 🤔

denkristoffer: Sounds very likely 😞

mshaaban0 (commented):
Works very well, thanks.

Did we tryout https://www.npmjs.com/package/inquirer-autocomplete-prompt for searching through the spaces list?

massao:
> Works very well, thanks.
> 
> Did we tryout https://www.npmjs.com/package/inquirer-autocomplete-prompt for searching through the spaces list?

Will check it

mshaaban0 (approved):
Looks good 👍

lib/cmds/init.ts
```diff
@@ -80,6 +81,10 @@ export const init = async () => {
         content: initialContent
       })
     }
+  } else {
+    const space = await spaceUse({ context })
+    // Return space to be used.
+    console.log({ space })
```
denkristoffer, line 87: Is this ready for review?

massao: nice catch, missed that one

lib/cmds/space_cmds/use.ts
```diff
@@ -94,11 +111,9 @@ async function spaceUse({ context, spaceId, header }) {
 
   await storeRuntimeConfig()
 
-  showSuccess(space, activeEnvironmentId)
+  showSuccess(space, activeEnvironmentId as string)
```
marcolink, line 114: Out of curiosity, why the cast?

mshaaban0: Context hasn't been migrated to TS yet. I'll fix that in another PR.
`showSuccess` needs the envId to print. So types for it is correct, context is what needs love.

marcolink (approved):

PR #1596, opened 2022-08-18
feat(init): success message

denkristoffer:
Shows a success message after going through the `init` flow.

Screenshot 2022-08-18 at 14 51 01


lib/cmds/init/success.ts
```diff
@@ -0,0 +1,68 @@
+import { highlight } from 'cli-highlight'
```
mshaaban0, line 1: I'd like to see `chalk` imported here :) we neeeeeed colors to differentiate between output sections!
Try not to over do it though, let's keep it subtle. 🙏🏼 ❤️

lib/cmds/init.ts
```diff
@@ -76,6 +77,15 @@ export const init = async () => {
       })
     }
   }
+
+  // something something, get the flow type
+  // one of 'Contentful.js', 'GraphQL', 'REST API'
+  const connectionType = 'GraphQL'
```
mshaaban0, line 83: Yup, we need another prompt here to choose one of ....

mshaaban0 (commented):
I can't seem to get any of the success output when going through the init command 🤔
Screenshot 2022-08-19 at 16 55 57

lib/cmds/init/success.ts
```diff
@@ -0,0 +1,68 @@
...
+  const client = contentful.createClient({
+    space: "${spaceId}",
+    accessToken: "${accessToken}"
+  });
+  client
+    .getEntry("${entryId}")
+    .then(entry => console.log(entry))
+    .catch(err => console.log(err));
+    `
+    ),
```
mshaaban0, line 33: second parameter accepts `language: 'javascript'`, it should help with coloring these 

```js
highlight(graphQlExample, {
  language: "JavaScript",
})
```

lib/cmds/init.ts
denkristoffer: Wouldn't that come from previously in the flow?

mshaaban0 (approved):
Let's merge this and I'll follow up with putting the steps together 👍🏼

PR #1603, opened 2022-08-25
feat: upgrade node versions used by pkg to build standalone executables

damienxy:
## Description

`pkg`, which builds standalone executables of Contentful CLI for Windows, Mac, Linux, was still using Node 12. Upgraded to Node 16.


marcolink (approved):
🚢

PR #1606, opened 2022-08-30
test: remove unnecessary pre and post test scripts

mshaaban0:
## Summary

Since we already use the `tmp` folder to store `.contentfulrc.json` we don't need these scripts to do cleanups anymore

## Next PR

Rewrite documentation on how to run tests locally


marcolink (approved):
Looking good 👍

PR #1610, opened 2022-09-02
fix: add defaultLocale as option to space create command

mshaaban0:
## Summary

Contentful management js lib types was not showing `defaultLocale` as an available option in `createSpace` method.
That led to removing it in #1583  See https://github.com/contentful/contentful-cli/pull/1583#discussion_r957541315

This PR adds the options back `contentful space create --default-locale`


denkristoffer (approved):
👍 

Is the type fixed in `10.12.1`?

massao (approved):

PR #1613, opened 2022-09-05
chore: clean up login command messaging structure

mshaaban0:
## Summary

- Clean up and structure logging messages for `login` command
- Migrate `utils/async.js` to typescript

Screenshot 2022-09-06 at 15 08 14

Screenshot 2022-09-06 at 00 43 52

### DIFF
--- a/lib/cmds/login.ts
+++ b/lib/cmds/login.ts
@@ -54,12 +54,13 @@ export const login = async ({
     token = managementTokenFlag
   } else {
     if (managementToken) {
+      console.log(`You're already logged in!`)
+      await tokenInfo()
       console.log(
-        `You're already logged in! to logout type: ${chalk.green(
-          'contentful'
-        )} ${chalk.magenta('logout')}`
+        `To logout, type: ${chalk.green('contentful')} ${chalk.magenta(
+          'logout'
+        )}\n`
       )
-      tokenInfo()
       return managementToken
     }
 
--- a/lib/cmds/login.ts
+++ b/lib/cmds/login.ts
@@ -2,13 +2,12 @@ import open from 'open'
 import inquirer from 'inquirer'
 import chalk from 'chalk'
 
-import { getConfigPath, setContext, storeRuntimeConfig } from '../context'
+import { setContext, storeRuntimeConfig } from '../context'
 import { confirmation } from '../utils/actions'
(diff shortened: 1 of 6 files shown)


lib/cmds/login.ts
```diff
@@ -55,28 +54,25 @@ export const login = async ({
     token = managementTokenFlag
   } else {
     if (managementToken) {
-      log()
-      log(
-        `Looks like you already stored a management token on your system. ${chalk.dim(
-          `(Located at ${await getConfigPath()})`
-        )}`
+      console.log(
+        `You're already logged in! to logout type: ${chalk.green(
```
denkristoffer, line 58:
```suggestion
        `You're already logged in! To logout type: ${chalk.green(
```

lib/cmds/login.ts
```diff
@@ -55,28 +54,25 @@ export const login = async ({
     token = managementTokenFlag
   } else {
     if (managementToken) {
-      log()
-      log(
-        `Looks like you already stored a management token on your system. ${chalk.dim(
-          `(Located at ${await getConfigPath()})`
-        )}`
+      console.log(
```
denkristoffer, line 57: Why did we drop the `log` utility?

lib/cmds/login.ts
denkristoffer, line 58: Personally I prefer the old message with the "maybe you want to log out?" coming later, I think it's a good separation whereas the logout hint is a bit harder to spot here.

lib/cmds/login.ts
massao: The `log` utility is just a wrapper on the `console.log`.
The others utility have a bit more of logic/styling involved, but we could keep it as log for consistency 🤔 
https://github.com/contentful/contentful-cli/blob/master/lib/utils/log.js

mshaaban0: Yup, it's just a wrapper that confuses me with other utility loggers we have, overtime I have to click to see what it is, I'd say it's not worth it. Maybe in another PR I'll just do a quick find and replace

denkristoffer: I think that's okay but let's kill it everywhere then, yeah.

denkristoffer (approved):

PR #1614, opened 2022-09-05
feat: add select connection type step to init command

mshaaban0:
## Summary

Add connection type selector to init command

Screenshot 2022-09-06 at 15 11 28

### DIFF
--- a/lib/cmds/init/success.ts
+++ b/lib/cmds/init/success.ts
@@ -62,5 +62,5 @@ export default async function success({
     )
   }
 
-  console.log(`Share your feedback at ${FEEDBACK_LINK}`)
+  console.log(`Share your feedback for this CLI at ${FEEDBACK_LINK}`)
 }
--- a/lib/cmds/init/success.ts
+++ b/lib/cmds/init/success.ts
@@ -29,7 +29,8 @@ export default async function success({
     .getEntry("${entryId}")
     .then(entry => console.log(entry))
     .catch(err => console.log(err));
-    `
+    `,
+      { language: 'JavaScript' }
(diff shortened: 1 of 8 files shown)


massao (approved):

lib/cmds/init/success.ts
```diff
@@ -61,8 +62,5 @@ export default async function success({
     )
   }
 
-  console.log(
-    'Do you want to give us feedback on the experience of using this command?'
-  )
-  console.log(`Share your thoughts at ${FEEDBACK_LINK}`)
+  console.log(`Share your feedback at ${FEEDBACK_LINK}`)
```
denkristoffer, line 65: I think it could be clearer this is feedback on the CLI init command only.

mshaaban0: This should be for the CLI as a whole, we need less noise with the code examples :)

denkristoffer: I agree with that.

denkristoffer:
```suggestion
  console.log(`Share your feedback for this CLI at ${FEEDBACK_LINK}`)
```
Maybe?

mshaaban0: Isn't the context quite obvious though?

denkristoffer: Maybe I'm dumb but I don't think it is to me 😆

mshaaban0: 😊 the whole flow will be reviewed by content writers anyway. Feel free to suggest a commit

PR #1632, opened 2022-09-19
chore: update copyright year

AbuHafsa:
## Summary

Small update of the copyright year.

## Motivation and Context

I noticed the copyright year wasn't changed since 2019 so I thought to do it before it gets forgotten again 😁

### DIFF
--- a/lib/cli.js
+++ b/lib/cli.js
@@ -25,7 +25,7 @@ yargs
     skipValidation: true
   })
   .version(false)
-  .epilog('Copyright 2019 Contentful')
+  .epilog('Copyright 2022 Contentful')
   .fail(function (msg, err, yargs) {
     if (err) throw err
     console.error(yargs.help())


AbuHafsa:
There are many places where the copyright year has to be changed. I will submit a general solution (that I am putting implementing) to this instead of this single change.

The solution I thought of, is make `lib/copyright.json` that looks like `{ "year": 2022, "company": "Contentful" }` and use an inline string that looks like `Copyright ${year} ${company}` in the epilog or any other place.

mshaaban0:
Hi @AbuHafsa Thanks a lot for your contribution. This will be addressed in the next beta version, see https://github.com/contentful/contentful-cli/pull/1486

AbuHafsa:
Hi @mshaaban0 Thank you for letting me know. I will close this PR then as it is not necessary.

PR #1633, opened 2022-09-19
fix: inject management token in context

mshaaban0:
## Summary

Inject management token to context before passing it to space import, use, or create. This to fix:

Screenshot 2022-09-19 at 16 16 27

### DIFF
--- a/lib/cmds/init.ts
+++ b/lib/cmds/init.ts
@@ -30,7 +30,10 @@ export const init = async () => {
     context.managementToken = await login({ context })
   }
 
-  if (!context.managementToken) return
+  if (!context.managementToken)
+    return console.log(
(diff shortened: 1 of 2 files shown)


lib/cmds/init.ts
```diff
@@ -25,13 +25,12 @@ export const init = async () => {
...
   const context = await getContext()
-  let managementToken = context.managementToken
 
   if (!context.managementToken) {
-    managementToken = await login({ context })
+    context.managementToken = await login({ context })
   }
 
-  if (!managementToken) return
+  if (!context.managementToken) return
```
denkristoffer, line 33: Should we show a message in this case?

denkristoffer (approved):
Nice!

PR #1636, opened 2022-09-20
feat: fetch and display CDA api preview key in code examples

mshaaban0:
## Summary

Previously we were displaying CMA access token in CDA requests code example. Instead we should display CDA access token. This PR does the following:

* Fetch or create CDA preview api access token
* Abstract space creation flow to it's own function for better readability 

## Todos Next PRs

- Write e2e tests for init command
- Try out https://github.com/highlightjs/highlight.js#nodejs-on-the-server instead of the old cli-highlights, fix and test code examples

Screenshot 2022-09-20 at 16 54 34

### DIFF
--- a/lib/cmds/init.ts
+++ b/lib/cmds/init.ts
@@ -5,11 +5,9 @@ import greetings from './init/greetings'
 import success from './init/success'
 import { getContext } from '../context'
 import { login } from './login'
-import { spaceCreate } from './space_cmds/create'
-import { importSpace } from './space_cmds/import'
-import initialContent from './init/content.json'
-import { spaceUse } from './space_cmds/use'
 import chalk from 'chalk'
+import { getPreviewApiKey } from './init/apikey'
+import { getSpace } from './init/space'
 
 export const command = 'init'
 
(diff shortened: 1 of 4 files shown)


denkristoffer:
Aren't we locking the user out of a lot of contentful.js functionality if their API key is for the CDA and not the CMA?

mshaaban0:
> Aren't we locking the user out of a lot of contentful.js functionality if their API key is for the CDA and not the CMA?

@denkristoffer I'm don't think that's the case, if so I'm not aware of that information, my understanding is `contentful.js` is meant to talk only to the delivery api, hence it only needs the CDA access token. https://contentful.github.io/contentful.js/contentful/9.1.18/contentful.html#.createClient

marcolink:
> Aren't we locking the user out of a lot of contentful.js functionality if their API key is for the CDA and not the CMA?

`contentful.js` is only interacting with the CDA/CPA - no win by using a CMA token here 🙃

denkristoffer (approved):
🤯

massao (approved):

marcolink (approved):

PR #1705, opened 2022-11-14
fix: update contentful-migration packge

damienxy:
## Summary

Updating the contentful-migration dependency to the newest version (`4.12.2`)


marcolink (approved):
Is dependabot still working?

PR #1708, opened 2022-11-18
feat: add options to skip or strip tags

veu:
## Summary

This change adds two new options to the export:

--skip-tags – Skips fetching tags
--strip-tags – Remove tags from entries and assets

Together they allow to remove all tag information from the export.

## Description

The functionality is implemented in https://github.com/contentful/contentful-export/pull/1191. This PR makes the new options visible in the CLI.


marcolink (approved):
🚢

PR #1805, opened 2023-01-30
Feat/app installation check [SDK-2440]

andreascful:
## Summary

Add a utils function that checks if a specified app is installed on an environment.

### DIFF
--- a/lib/utils/app-installation.ts
+++ b/lib/utils/app-installation.ts
@@ -1,0 +1,29 @@
+import { ClientAPI } from 'contentful-management'
+
+/**
+ * Checks if a specified app is installed in an environment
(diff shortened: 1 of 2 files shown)


lib/utils/app-installation.ts
```diff
@@ -0,0 +1,29 @@
+import { ClientAPI } from 'contentful-management'
+
+/**
+ * Checks if a specified app is installed in an environment
+ */
+async function appIsInstalled(
```
marcolink, line 6: I think `IsAppInstalled`. would more align with our naming convention

Vinz93 (approved):
Muy bien 🙂

PR #1806, opened 2023-01-30
Feat: Add `merge` command and initial `export` subcommand 

crissto:
## Summary

Adds the scaffold to add commands to the `merge` subcommand and the initial `merge export` command with source and target envs

### DIFF
--- a/lib/cmds/merge_cmds/export.ts
+++ b/lib/cmds/merge_cmds/export.ts
@@ -1,0 +1,49 @@
+import type { Argv } from 'yargs'
+
+module.exports.command = 'export'
+
+module.exports.desc = 'Export diff between two environments as a migration'
+
+module.exports.builder = (yargs: Argv) => {
+  return yargs
+    .usage('Usage: contentful merge export')
+    .option('source-environment-id', {
+      alias: 's',
+      type: 'string',
+      demand: true,
+      describe: 'Source environment id'
+    })
+    .option('target-environment-id', {
+      alias: 't',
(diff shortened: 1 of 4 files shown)


lib/cmds/merge_cmds/export.ts
```diff
@@ -0,0 +1,49 @@
+import type { Argv } from 'yargs'
+
+module.exports.command = 'export'
+
+module.exports.desc = 'Export diff between two environments as a migration'
+
+module.exports.builder = (yargs: Argv) => {
+  return yargs
+    .usage('Usage: contentful merge export')
+    .option('source-environment-id', {
```
marcolink, line 10: Do we also need the space id?

crissto: Yes! I didn't add it because it gets automatically asked for the context here https://github.com/contentful/contentful-cli/blob/14beebf03bebaea4a42ce8f014842034fa514cfe/lib/utils/middlewares.js#L67. 
The context adds the `--space-id` for most of the commands. It also allows the user to set the context before by running `contentful use <space-id>`.
I could add it as an option for clarity though

marcolink: We should provide a non-blocking way of executing the command, also if no rc file is present.

crissto:
> Wondering if we have a simple way to test existence of the commands
We actually have that. There are snapshot tests. Right now it tests for the merge command to exists but when export get's done it will have the check for export too

PR #1809, opened 2023-01-31
build(deps): remove snake cmd to remove blessed.

hwartig:
Removing the snake command could be considered a breaking change. Given that it was never a documented command but more of an easter egg we don't consider removing it a breaking change.

If your production setup relies on it consider replacing it with [CLI-snake](https://replit.com/@UniqueOstrich18/CLI-Snake).


lib/config.js
```diff
@@ -2,7 +2,6 @@ module.exports = {
   noAuthNeeded: [
     'login',
     'guide',
-    'snake',
```
mshaaban0, line 5: Nice, since this is a "breaking change", can we please have it as part of V2 against the `beta` branch?

hwartig: Have you seen my comment about it not being a breaking change because it wasn't part of the official api? Do you think it will actually break someones setup?

hwartig: Ok you convinced me I'll change merge base to beta. Don't want you to need to go through beta rebase again.

mshaaban0 (approved):

PR #1815, opened 2023-02-03
Feat: add app installation before export [SDK-2441]

crissto:
## Summary
For the upcoming commands we need to check if the merge app is installed in the user provided environments.

## Description

Adds the flow check if app is installed in the env and if it isn't and the users confirms it goes ahead and installs it


lib/utils/app-installation.ts
```diff
@@ -32,4 +32,29 @@ async function isAppInstalled(
...
+    environmentId: string | string[]
+    appId: string
+  }
+): Promise<void> {
+  const environments = Array.isArray(environmentId)
+    ? environmentId
+    : [environmentId]
+
+  for (const environmentId of environments) {
+    await client.rawRequest({
```
andreascful, line 52: I am wondering if this function could be used instead of the rawRequest. https://github.com/contentful/contentful-management.js/blob/97adca455dd85a8552ae6510ff560e7bb058b07f/lib/entities/app-installation.ts#L47

test/integration/cmds/merge/export.test.ts
```diff
@@ -0,0 +1,85 @@
...
+  return nixt({ newlines: true }).cwd(bin).base('./contentful.js ').clone()
+}
+
+type Result = {
+  stderr: string
+  stdout: string
+}
+
+describe('merge export snapshots', () => {
+  it('show the help properly', done => {
```
andreascful, line 16:
```suggestion
  it('shows the help properly', done => {
```

test/unit/cmds/merge_cmds/export.test.ts
```diff
@@ -0,0 +1,65 @@
...
+
+  beforeEach(() => {
+    isAppInstalled.mockClear()
+    installApp.mockClear()
+  })
+
+  it('stops early if both env have the app installed', async () => {
+    isAppInstalled.mockResolvedValue(true)
+
+    const ret = await checkAndInstallAppInEnvironments(
```
andreascful, line 21: Just curious what `ret` stands for :)

test/unit/cmds/merge_cmds/export.test.ts
```diff
@@ -0,0 +1,65 @@
...
+      ['source', 'target'],
+      'app-id',
+      false
+    )
+
+    expect(ret).toBe(true)
+    expect(isAppInstalled).toHaveBeenCalledTimes(2)
+  })
+
+  it('installs app to both apps if none of them have it installed', async () => {
```
andreascful, line 33:
```suggestion
  it('installs app to both envs if none of them have it installed', async () => {
```

andreascful (approved):
Great PR!

lib/utils/app-installation.ts
crissto: That function seems to do an upsert and it seems like you need an existing appInstallation in the environment which we don't have. 
Also, as we are not using the plain client we would need to fetch space and env before updating, making the request count a lot higher.

test/integration/cmds/merge/__snapshots__/export.test.ts.snap
```diff
@@ -0,0 +1,14 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`merge export snapshots show the help properly: help data is incorrect 1`] = `
+"Usage: contentful merge export
+
+Options:
+  -h, --help                   Show help                               [boolean]
+  --source-environment-id, -s  Source environment id         [string] [required]
+  --target-environment-id, -t  Target environment id         [string] [required]
+  --yes, -y                    Confirm merge app installation without prompt
```
damienxy, line 10:
```suggestion
  --yes, -y                    Confirm Merge app installation without prompt
```

lib/cmds/merge_cmds/export.ts
```diff
@@ -19,6 +25,10 @@ module.exports.builder = (yargs: Argv) => {
       demand: true,
       describe: 'Target environment id'
     })
+    .option('yes', {
+      alias: 'y',
+      describe: 'Confirm merge app installation without prompt'
```
damienxy, line 30:
```suggestion
      describe: 'Confirm Merge app installation without prompt'
```

lib/utils/app-installation.ts
```diff
@@ -32,4 +34,124 @@ async function isAppInstalled(
...
+    `The Merge app is not installed in the environment with id: ${environmentId}`
+  )
+
+  const userConfirmation = await confirmation(
+    `Do you want to install the Merge app in the environment with id: ${environmentId}`
+  )
+
+  if (userConfirmation) {
+    return false
+  }
```
damienxy, line 80: Should this be checking for `!userConfirmation`?

crissto: Yes! When testing things for the Show & Tell I came across this.

crissto:
Closed in favor of #1832

PR #1819, opened 2023-02-07
Chore: merge beta keeping old name []

crissto:
## Summary
Due to pending work needing to be done this PR modifies the name back to `contentful-cli` so we can release a new major with all the new changes that live in beta and weren't released.
This commit [contentful/contentful-cli@`8904c1a` (#1819)](https://github.com/contentful/contentful-cli/pull/1819/commits/8904c1a14ddf334ae9c8d9e522e98eb4dd61d066) should force a new relase by semantic-release

## Description

Branches out of beta only adding a new commit to change the package name back

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -116,4 +116,5 @@ workflows:
           filters:
             branches:
               only:
+                - master
                 - beta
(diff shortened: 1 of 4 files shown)


mshaaban0 (commented):
Also let's make sure we have "BREAKING CHANGE" in the commit message before merging

mshaaban0 (approved):
Looks good

PR #1851, opened 2023-02-24
feat: add `merge export` command [SDK-2501]

crissto:
## Summary
It adds the `merge export` command that replicates the functionality from the Merge App where given a source and a target environment it creates the migration script.

## Description
Adds config needed for calling the app actions and completes the export command.

## Motivation and Context
Enabling users to create content-type migration between envs without the use of a UI


lib/cmds/merge_cmds/export.ts
```diff
@@ -50,13 +61,62 @@ interface ExportMigrationOptions {
...
+    changesetRef = await callCreateChangeset({
+      api,
+      appDefinitionId,
+      appActionId: createChangesetActionId,
+      parameters: {
+        sourceEnvironmentId,
+        targetEnvironmentId
+      },
+      spaceId,
+      environmentId: targetEnvironmentId
```
ruderngespra, line 93: Why is this always the `targetEnvironmentId` and not the default environment id picked up by the config?

lib/utils/app-actions.ts
```diff
@@ -0,0 +1,101 @@
...
+  const params = {
+    appDefinitionId,
+    appActionId,
+    ...(environmentId ? { environmentId } : {}),
+    ...(spaceId ? { spaceId } : {})
+  }
+
+  const createResponse = await api.appActionCall.create(params, { parameters })
+
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
```
ruderngespra, line 51: 🙈 lol

lib/cmds/merge_cmds/export.ts
ruderngespra: As discussed, I would suggest adding a comment (to the code) explaining why we use the targetEnvironment here.

ruderngespra (approved):
Lgtm, I'd suggest to add one comment on the `targetEnvironmentId` selection as context environment, as this may  be confusing for anyone afterwards.

ruderngespra:
Just saw that there are some typing issues in the unit tests related to the `app-action-utils` bump, will look into it :/

ruderngespra (approved):

PR #1854, opened 2023-02-27
chore: increase jest timeout to fix integration test pipeline

marcolink:
Increase global jest timeout as we saw several integration test run failing.


mayakarabula (approved):

andreascful (approved):
Did it fail because of waiting for envs?

marcolink:
> Did it fail because of waiting for envs?

it fails because space creation takes > 5s

PR #1879, opened 2023-03-06
fix: [ATOOLS-1533] use yaml extension for catalog-info

omasopust:
## Summary

the extension matters as both Backstage and CI Alerts are only looking for `.yaml`

## Todos

- [x] Implemented feature
- [ ] Feature with pending implementation


crissto (approved):
Thanks!!

Issue #1833, opened 2023-02-13
Standalone CLI going away?


99littlebugs:
One of the [breaking changes](https://github.com/contentful/contentful-cli/releases/tag/v2.0.0) for the CLI version 2.0.0 was listed as `Removed Linux and Mac OSX standalone packages`. Is that the go-forward strategy? We've been using the standalone CLI downloads and are sad to see them being removed. 

## Context
We like to run migrations through the CLI and have a requirement to treat all content model updates like code (put them through code review and the like). Therefore, we have installed the contentful CLI on our production server and use it to run migrations (with production credentials) after the migrations are deployed. Our servers do not currently have a node installed and we'd prefer to keep it that way.

0xStarcat:
same issue here, my company has a python monolith and uses the executable to run migrations. Would like to install the linux standlone CLI into our docker image without needing to install npx in the image. 

The [installation docs](https://www.contentful.com/developers/docs/tutorials/cli/installation/) still point to the latest release for the linux executable, but it's no longer there in version 2+. 

Is there an alternative way to compile the executable from the source code provided in version 2+ that can be documented? Or some explanation why it was removed or plans to reintroduce it?

thanks!

marcolink:
Hello @99littlebugs @0xStarcat ,
We have re-added standalone builds on your request: 
https://github.com/contentful/contentful-cli/releases/tag/v2.2.3

PR #1893, opened 2023-03-13
feat: standalone CLI [SDK-2554]

marcolink:
## Summary

Bring back standalone builds upon request.

Solves [#1833](https://github.com/contentful/contentful-cli/issues/1833)

**Important:**
We have to add `axios` CJS file explicitly as asset, because `pkg`. can't handle ESM modules.
https://github.com/vercel/pkg/issues/1822

(If this still causes problems, we have to downgrade to < 1)


damienxy (approved):
LGTM, I would also link the issue in the description.

PR #1919, opened 2023-03-22
chore: use different aliases for merge command

ruderngespra:


test/integration/cmds/merge/export.test.ts
```diff
@@ -39,7 +39,7 @@ describe('merge export command args validation', () => {
 
   it('should exit 1 when no source environment', done => {
     app()
-      .run('merge export --t target')
+      .run('merge export --se target')
```
damienxy, line 42: I think this should be `--te` (test title says "no source environment")

test/integration/cmds/merge/export.test.ts
```diff
@@ -50,7 +50,7 @@ describe('merge export command args validation', () => {
 
   it('should exit 1 when no target environment', done => {
     app()
-      .run('merge export --s source')
+      .run('merge export --ta source')
```
damienxy, line 53: I think this should be `--se` (test title says "no target environment")

damienxy (approved):
Nice! 
Should we do the change [on master](https://github.com/contentful/contentful-cli/blob/master/lib/cmds/merge_cmds/export.ts) as well or can we wait until we merge?

test/integration/cmds/merge/export.test.ts
ruderngespra: 👍🏻

ruderngespra:
> Nice! Should we do the change [on master](https://github.com/contentful/contentful-cli/blob/master/lib/cmds/merge_cmds/export.ts) as well or can we wait until we merge?

I think this can wait.

PR #1935, opened 2023-03-27
feat(merge): use new cmd param alias

marcolink:
### Switch to parameter aliases for merge command

Additional changes:

- introduce Readme files for merge commands
- extract shared merge command setup logic into a helper function.


andreascful (approved):
Really nice! 👍

PR #1990, opened 2023-04-14
fix: better errors on merge cli commands []

crissto:
Right now when we have a `PollTimeout` error the output is not the best. This fixes it and sets more user friendly messages.

### DIFF
--- a/lib/cmds/merge_cmds/export.ts
+++ b/lib/cmds/merge_cmds/export.ts
@@ -8,10 +8,9 @@ import {
 import { getAppActionId, type Host } from '../../utils/app-actions-config'
 import { handleAsyncError as handle } from '../../utils/async'
 import { ensureDir, getPath, writeFileP } from '../../utils/fs'
-import { error, success } from '../../utils/log'
+import { success } from '../../utils/log'
 import { prepareMergeCommand } from '../../utils/merge/prepare-merge-command'
 import { MergeContext } from '../../utils/merge/types'
-import { errorEmoji } from '../../utils/emojis'
 import { mergeErrors } from '../../utils/merge/errors'
 
 module.exports.command = 'export'
@@ -88,8 +87,9 @@ export const callExportAppAction = async ({
       environmentId: targetEnvironmentId
     })
   } catch (e) {
-    throw new Error('Changeset could not be created.')
+    throw new Error(mergeErrors['ErrorInDiffCreation'])
   }
+
   try {
     const { migration } = await getExportMigration({
       api,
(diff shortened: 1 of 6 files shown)


lib/cmds/merge_cmds/show.ts
```diff
@@ -79,27 +77,44 @@ export const getChangesetAndTargetContentType = async ({
...
+    )
+  }
 
-  const { result } = appActionResult
+  const { result } = appActionResult.value
 
   if (isResultWithError(result)) {
+    if (result.errorMessage === 'PollTimeout') {
+      throw new Error(
+        `${errorEmoji} The migration took too long to generate. Please try again.`
```
andreascful, line 106: Since this message is used in multiple places I wonder if we should store in centrally somewhere where it can be used for any PollTimeout error

lib/cmds/merge_cmds/export.ts
```diff
@@ -142,6 +152,10 @@ const exportEnvironmentMigration = async ({
       targetEnvironmentId,
       spaceId: activeSpaceId
     })
+
+    if (!migration) {
+      throw new Error('Migration could not be exported.')
```
damienxy, line 157: This is the same error message as above - should we put it in a const so it's reusable?

lib/cmds/merge_cmds/export.ts
```diff
@@ -142,17 +152,21 @@ const exportEnvironmentMigration = async ({
...
+    if (!migration) {
+      throw new Error(mergeErrors['MigrationCouldNotBeExported'])
+    }
   } catch (e) {
     if (e instanceof Error) {
       throw e.message
     }
 
-    throw new Error('Migration could not be exported.')
+    throw new Error(mergeErrors['MigrationCouldNotBeExported'])
```
damienxy, line 164: All of this is already handled by callExportAppAction (we catch and rethrow with new error messages there), so I think here we can just propagate the errors, no?

lib/cmds/merge_cmds/show.ts
```diff
@@ -49,14 +46,16 @@ export const getChangesetAndTargetContentType = async ({
   host,
   appDefinitionId,
   sourceEnvironmentId,
-  targetEnvironmentId
+  targetEnvironmentId,
+  timeout
```
damienxy, line 50: `timeout` is not used, should we remove it?

marcolink (approved):

lib/cmds/merge_cmds/export.ts
```diff
@@ -138,35 +136,24 @@ const exportEnvironmentMigration = async ({
...
-      createChangesetActionId: getAppActionId('create-changeset', host as Host),
-      exportActionId: getAppActionId('export-changeset', host as Host),
-      sourceEnvironmentId,
-      targetEnvironmentId,
-      spaceId: activeSpaceId
-    })
-
-    if (!migration) {
-      throw new Error(mergeErrors['MigrationCouldNotBeExported'])
-    }
```
damienxy: Is this something that can happen if no error is thrown above? In that case we still need this check

crissto: We check if whats returned looks good here https://github.com/contentful/contentful-cli/blob/d4296c3e11e85ca12f481ed13a2861f37a9e1415/lib/utils/app-actions.ts#L100 so if migration doesn't exist at this point an error would have already been thrown and that makes this check redundant

damienxy (approved):

PR #2014, opened 2023-04-27
feat: only show content type changes in merge show command [PHX-2514]

damienxy:
Temporary filter to only show content type changes, while editor interfaces are not supported yet.


Vinz93 (approved):
🛳️ 🎉

PR #2134, opened 2023-06-28
fix: npm audit fix []

mayakarabula:


marcolink:
@mayakarabula what is the outcome of running the audit fix command? can we maybe add a screenshot what changed? Did we do it because dependabot didn't fix a vulnerability?

mayakarabula:
@marcolink I have updated description with more information

PR #2244, opened 2023-08-31
feat: show moves in merge show command output [PHX-2582]

damienxy:
## Summary

During the `merge show` command, move operations are now also displayed.

## Description
For field moves, we indicate the direction (up or down). 
For more deeply nested moves, we indicate that the order has changed and list the (nested) properties.

## Screenshot:

The screenshot shows the following cases: 
- nested property has order changes
- deeply nested property has order changes
- field was moved up
- field was moved down

image


marcolink (approved):
It's beautiful 🌷

PR #2297, opened 2023-09-25
chore: drop unsupported node versions and adds node v18 support [EXT-4785]

jjolton-contentful:
- swaps to m1 mac for osx circleci
- adds support for node v18.x


package.json
```diff
@@ -120,7 +120,7 @@
     "yargs": "~13.3.2"
   },
   "engines": {
-    "node": ">=16"
+    "node": ">=20.6.1"
```
t-col, line 123: I think we should use node 18 and above as the engine cuz 18 will still be supported in maintenance LTS for another year or so, and customers may not be on 20 yet. I realized that `yarn` will actually error on install if your version is out of date with a package's engine, but `npm` will just give you a warning

I think we should keep using 20 in the nvmrc and CI tho

.circleci/config.yml
```diff
@@ -50,8 +50,10 @@ jobs:
           path: ./reports/e2e-tests.xml
   osx:
     macos:
-      xcode: 12.5.1
+      xcode: 14.3.1
+    resource_class: macos.m1.medium.gen1
     steps:
+      - run: nvm install 20.6.1 && nvm alias default 20.6.1
```
t-col, line 56: I wonder if installing node this way is messing with the `pkg` library that's building the executables

jjolton-contentful:
jsonpath (a dependency of bfj, which is used by this repo and contentful-batch-libs) is incompatible with pkg, and thus we need to force bfj to use the browserified jsonpath.min instead - see https://github.com/dchester/jsonpath/issues/96

package.json
```diff
@@ -70,9 +71,10 @@
     "lint-staged": "^14.0.0",
     "nixt": "^0.5.1",
     "nyc": "^15.0.0",
+    "patch-package": "^8.0.0",
     "pkg": "^5.8.1",
     "rimraf": "^3.0.2",
-    "semantic-release": "^22.0.5",
+    "semantic-release": "^21.0.0",
```
andreascful, line 77: Was this downgrade done on purpose?
If so you might want to pin to that version.

patches/bfj+7.1.0.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/node_modules/bfj/src/match.js b/node_modules/bfj/src/match.js
```
andreascful, line 1: Really interesting investigation overall! 
I am not sure if I am the biggest fan of the patch though because patching usually is hard to understand for a reader of the code. The underlying lib might also make changes in the future that would make this workaround break since we are basically relying on an implementation detail.

However, the alternative of permanently pegging to an old version also doesn't seem too appealing (missing out on any updates, esp. security). 
So overall I'd suggest adding some comments explaining the context of the patch to the patch so that later we decrease the friction of understanding it as much as possible. Wdyt?

andreascful (changes requested):
⬆️

package.json
jjolton-contentful: Using the latest semantic-release (22.0.5), it prints these warnings: 

> npm WARN EBADENGINE Unsupported engine {
npm WARN EBADENGINE   package: 'semantic-release@22.0.5',
npm WARN EBADENGINE   required: { node: '^18.17 || >=20.6.1' },
npm WARN EBADENGINE   current: { node: 'v18.15.0', npm: '9.5.0' }
npm WARN EBADENGINE }
npm WARN EBADENGINE Unsupported engine {
npm WARN EBADENGINE   package: '@semantic-release/commit-analyzer@11.0.0',
npm WARN EBADENGINE   required: { node: '^18.17 || >=20.6.1' },
npm WARN EBADENGINE   current: { node: 'v18.15.0', npm: '9.5.0' }
npm WARN EBADENGINE }
npm WARN EBADENGINE Unsupported engine {
npm WARN EBADENGINE   package: '@semantic-release/npm@11.0.0',
npm WARN EBADENGINE   required: { node: '^18.17 || >=20' },
npm WARN EBADENGINE   current: { node: 'v18.15.0', npm: '9.5.0' }
npm WARN EBADENGINE }
npm WARN EBADENGINE Unsupported engine {
npm WARN EBADENGINE   package: '@semantic-release/release-notes-generator@12.0.0',
npm WARN EBADENGINE   required: { node: '^18.17 || >=20.6.1' },
npm WARN EBADENGINE   current: { node: 'v18.15.0', npm: '9.5.0' }
npm WARN EBADENGINE }
npm WARN EBADENGINE Unsupported engine {
npm WARN EBADENGINE   package: 'env-ci@10.0.0',
npm WARN EBADENGINE   required: { node: '^18.17 || >=20.6.1' },
npm WARN EBADENGINE   current: { node: 'v18.15.0', npm: '9.5.0' }
npm WARN EBADENGINE }
npm WARN EBADENGINE Unsupported engine {
npm WARN EBADENGINE   package: 'npm@10.2.0',
npm WARN EBADENGINE   required: { node: '^18.17.0 || >=20.5.0' },
npm WARN EBADENGINE   current: { node: 'v18.15.0', npm: '9.5.0' }
npm WARN EBADENGINE }

Unfortunately pkg only supports node 18.15.0 right now (see https://github.com/vercel/pkg-fetch/blob/main/patches/patches.json), but from what I can tell, everything works fine even with those warnings so I went ahead and reverted this change

patches/bfj+7.1.0.patch
jjolton-contentful: Great suggestion, I added a readme to the patches folder to consolidate the story on this. I am really not a big fan of the patch either :/ but totally agree that from a security/stability perspective it seems like a better solution than pinning to an old version.

Definitely crossing my fingers that a fix makes it into either bfj or jsonpath soon

andreascful (approved):
Thank you for the changes. I think this is the best we can do right now!

PR #2306, opened 2023-09-26
chore(): update xcode version

andreascful:
Updating the Xcode version & adding a Gen2 resource class because CircleCI is deprecating MacOS intel Gen1.
This one [will be deprecated in January 2024](https://discuss.circleci.com/t/macos-intel-support-deprecation-in-january-2024/48718), so by then we'll have to switch to a silicon mac resource class.


ruderngespra (approved):
lgtm :)

PR #2330, opened 2023-10-09
chore: fix CI release step [PHX-2905]

marcolink:
~~Downgrading `semantic-release` and blocking any further updates until we migrate to `node@20` since `semantic-release@22` requires `node@20`~~

**Update:**
Thanks @damienxy for pointing out the actual version conflict. 
I changed the node image version to please the requirements from semantic-release.

**CI Release step error**
``` 
[semantic-release]: node version ^18.17 || >=20.6.1 is required. Found v18.15.0.
```


damienxy:
I think any node 18 version beyond node 18.17 should work, no? Current LTS is 18.18

Issue #2348, opened 2023-10-13
Error: Cannot find module esprima


wkruse:
## Expected Behavior

Running standalone `contentful space export ...` should work.

Following standalone Linux versions **work**

- v2.2.24
- v2.6.46
- v2.7.0
- v2.8.9

## Actual Behavior

Running standalone `contentful space export ...` results in following exception:

```
pkg/prelude/bootstrap.js:1872
      throw error;
      ^

Error: Cannot find module 'esprima'
Require stack:
- aesprim
1) If you want to compile the package/file into executable, please pay attention to compilation warnings and specify a literal in 'require' call. 2) If you don't want to compile the package/file into executable and want to 'require' it from filesystem (likely plugin), specify an absolute path in 'require' call using process.cwd() or process.execPath.
    at Module._resolveFilename (node:internal/modules/cjs/loader:946:15)
    at Function._resolveFilename (pkg/prelude/bootstrap.js:1951:46)
    at Function.resolve (node:internal/modules/cjs/helpers:108:19)
    at Object.<anonymous> (/snapshot/project/node_modules/jsonpath/lib/aesprim.js:4:20)
    at Module._compile (pkg/prelude/bootstrap.js:1926:22)
    at Object.<anonymous> (/snapshot/project/node_modules/jsonpath/lib/aesprim.js:15:11)
    at Module._compile (pkg/prelude/bootstrap.js:1926:22)
    at Module._extensions..js (node:internal/modules/cjs/loader:1166:10)
    at Module.load (node:internal/modules/cjs/loader:988:32)
    at Module._load (node:internal/modules/cjs/loader:834:12)
    at Module.require (node:internal/modules/cjs/loader:1012:19)
    at Module.require (pkg/prelude/bootstrap.js:1851:31)
    at require (node:internal/modules/cjs/helpers:102:18)
    at Object.<anonymous> (/snapshot/project/node_modules/jsonpath/lib/handlers.js:1:77)
    at Module._compile (pkg/prelude/bootstrap.js:1926:22)
    at Module._extensions..js (node:internal/modules/cjs/loader:1166:10) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [ 'aesprim' ],
  pkg: true
}

Node.js v18.5.0
```

Following versions are broken

- v2.8.10
- v2.8.27
- v3.0.5

## Possible Solution

Downgrading to a version that works, like **v2.8.9**.

## Environment

[...]

PR #2381, opened 2023-10-26
fix: remove bfj in favor of straightforward readFile+parse [EXT-4785]

jjolton-contentful:
## Summary

This PR is introduced to fix https://github.com/contentful/contentful-cli/issues/2348. Bfj does not play nice with pkg, so swapping for a standard readFile/parse allows us to remove the nasty postinstall logic that was used to remedy the incompatibility. See https://github.com/contentful/contentful-cli/pull/2297 for history around that. 

Since we were already loading the entire content into memory, I'm not seeing any specific reason forcing us to use bfj here.


mgoudy91 (approved):
Not seeing any issues with this

PR #2385, opened 2023-10-30
fix: patch contentful-export/bfj because it is incompatible with pkg []

jjolton-contentful:
## Summary

This PR is introduced to fix https://github.com/contentful/contentful-cli/issues/2348

## Motivation and Context

See https://github.com/contentful/contentful-cli/pull/2297 for additional history


t-col (approved):
don't have a ton of context here, but lgtm generally

jjolton-contentful:
> don't have a ton of context here, but lgtm generally

thanks for TAL. I should have caught this one earlier, I thought we had gotten rid of bfj entirely with https://github.com/contentful/contentful-cli/pull/2381, didn't catch that contentful-export is using it as well

PR #2420, opened 2023-11-16
chore: change ownership to developer-experience [PHX-2975]

ruderngespra:
update ownership information from `phoenix` to `developer-experience`

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,2 +1,4 @@
-*   @contentful/team-phoenix
+* @contentful/team-developer-experience
+
 lib/cmds/init @contentful/team-core-experience
+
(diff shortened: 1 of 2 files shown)


.github/CODEOWNERS
```diff
@@ -1,2 +1 @@
-*   @contentful/team-phoenix
-lib/cmds/init @contentful/team-core-experience
```
mayakarabula, line 2: are we sure we can remove it? 👀

marcolink: I think we should keep the path based ownership for the core-experience team.

ruderngespra: adjusted it

mayakarabula (approved):

PR #2586, opened 2024-02-20
docs: add docs for configuring the CLI for EU DR

t-col:
demonstrate how to configure the CLI for EU usage


mgoudy91 (approved):
good catch

PR #2677, opened 2024-06-24
fix: bump contentful-migration from 4.20.2 to 4.21.0 [SPA-2047]

Chaoste:
Ship the fix for experiences that was introduced in https://github.com/contentful/contentful-migration/pull/1320


t-col (approved):

t-col:
changed the title so semantic-release will actually release a new version

t-col (approved):

PR #2684, opened 2024-07-11
Upgrade contentful-import, add ability to skip entries

umazalakain:
Upgrades contentful-import and uses its ability to skip entries.

### DIFF
--- a/lib/cmds/space_cmds/import.ts
+++ b/lib/cmds/space_cmds/import.ts
@@ -54,6 +54,18 @@ export const builder = (yargs: Argv) => {
       type: 'boolean',
       default: false
     })
+    .option('skip-asset-updates', {
+      describe:
+        'Skips asset updates. Skips updating existing assets, only new ones are created. Useful for incremental imports',
+      type: 'boolean',
+      default: false
+    })
+    .option('skip-content-updates', {
+      describe:
+        'Skips content updates. Skips updating existing content, only new content are created. Useful for incremental imports',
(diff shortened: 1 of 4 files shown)


lib/cmds/space_cmds/import.ts
```diff
@@ -54,6 +54,18 @@ export const builder = (yargs: Argv) => {
...
     })
+    .option('skip-asset-updates', {
+      describe:
+        'Skips asset updates. Skips updating existing assets, only new ones are created. Useful for incremental imports',
+      type: 'boolean',
+      default: false
+    })
+    .option('skip-content-updates', {
+      describe:
+        'Skips content updates. Skips updating existing content, only new content are created. Useful for incremental imports',
```
jjolton-contentful, line 65:
```suggestion
        'Skips content updates. Skips updating existing content, only new content is created. Useful for incremental imports',
```

jjolton-contentful:
thanks again for the contribution @umazalakain, so exciting to see this come together!!! it would be really awesome to include some tests with this change now that the new functionality is available

jjolton-contentful (changes requested):
Thanks again for the contribution @umazalakain, really so exciting to see all the pieces of this come together!!! It would be really awesome to include some tests with this change now that the new functionality is available. The existing tests ("should print help message" and "should exit 1 when no args" in test/integration/cmds/space/import.test.js) also require a tweak to accompany these changes

PR #2790, opened 2024-10-22
feat: organization export 

momenzadehvida:
Adds a new command to export organization entities. For now this includes only taxonomy data.

As a user, I should be able to run contentful organization export --organization-id <organization-id> and get the export of my taxonomy like:

`{ "taxonomy": { "concepts": [], "conceptSchemes": [] } }`

### DIFF
--- a/docs/organization/export/README.md
+++ b/docs/organization/export/README.md
@@ -7,7 +7,7 @@ This command helps you export organization data to a json file
 ```
 contentful organization export
 
-export a space data to a json file
+export an organization's data to a json file
 
 Options:
   -h, --help                Show help                                    [boolean]
--- a/docs/organization/export/README.md
+++ b/docs/organization/export/README.md
@@ -1,0 +1,36 @@
+# Contentful CLI - `organization export` command
+
+This command helps you export organization data to a json file
+
+## Usage
+
+```
+contentful organization export
+
+export a space data to a json file
+
+Options:
+  -h, --help                Show help                                    [boolean]
+  --organization-id         ID of organization                             [string]
+
+  --save-file               Save the export as a json file
+                                                         [boolean] [default: true]
+
+  --header                  Pass an additional HTTP Header
+  --output-file             Output file.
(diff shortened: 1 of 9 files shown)


lib/cmds/organization_cmds/export.ts
```diff
@@ -0,0 +1,130 @@
...
+    .option('header', {
+      alias: 'H',
+      type: 'string',
+      describe: 'Pass an additional HTTP Header'
+    })
+    .option('output-file', {
+      alias: 'o',
+      type: 'string',
+      describe:
+        'Output file. It defaults to ./migrations/<timestamp>-<organization-id>.json'
```
marcolink, line 40: In the current implementation, this doesn't hold true. We currently print to `stdout` if this flag is not provided.  Maybe we want to the the same as we do for `contentful-export` [where we explicitly define a flag](https://github.com/contentful/contentful-export/blob/main/lib/usageParams.js#L114) for persisting the export in a file (default will print to stdout). 

- If `save-file` is provided, we persist to a file
   - if no `output-file` is not provided, we fall back to a default file name

marcolink (commented):
Should we add a docs page for `organization export` [here](https://github.com/contentful/contentful-cli/tree/main/docs/organization)?

marcolink (approved):

docs/organization/export/README.md
```diff
@@ -0,0 +1,39 @@
+# Contentful CLI - `organization export` command
+
+This command helps you export organization data to a json file
+
+## Usage
+
+```
+contentful organization export
+
+export a space data to a json file
```
jjolton-contentful, line 10:
```suggestion
export an organization's data to a json file
```

could even be more specific that it is just taxonomy data rn to cut down on confusion, up to you

jjolton-contentful (approved):

marcolink (dismissed):

jjolton-contentful (approved):

PR #2797, opened 2024-10-28
feat: add taxonomy transform command [GROOT-1496]

mayakarabula:
## Summary

Introduce taxonomy transform command that allows to convert external CSV to contentful JSON. We also want to expose a transform taxonomy context.

To achieve that we expose context with Taxonomy class with helper methods `addConcept`, `getConcept`, `addConceptScheme`... retrieved objects have help methods like `setPrefLabel` with that users can more easily work with on transforming their concepts and schemes

## Description
This is the proposed taxonomy transform context shape
```typescript
interface TransformContext {
  csv: {
    // parses any CSV to a JSON
    parse:
      | (<T>(
          csvString: string,
          config?: Papa.ParseConfig
        ) => Papa.ParseResult<T>)
      | (<T>(
          file: File,
          config?: Papa.ParseConfig
        ) => Promise<Papa.ParseResult<T>>)
  }
  fs: {
    // reads any file from disc
    readFile: typeof readFileP
    cwd: typeof process.cwd
  }
  taxonomy: Taxonomy
}
```


docs/organization/taxonomy-transform/examples/example-transform.js
```diff
@@ -0,0 +1,57 @@
+module.exports = async function ({ csv, taxonomy, fs }) {
```
momenzadehvida, line 1: is this transform function works for example-data?

mayakarabula: It should, yes 😅  it's from the example script & data you shared

momenzadehvida: It was this one, https://github.com/contentful/contentful-cli/blob/7cc0633ee32e22d51495c1aad9b179efcfa1bbd6/docs/organization/taxonomy-transform/examples/example-spreadsheet.csv it looks different.

momenzadehvida (approved):
Great job! :tada:

docs/organization/taxonomy-transform/examples/example-data.csv
```diff
@@ -0,0 +1,30 @@
+animalia,animalia,A taxonomy focused on the categorization of various animals and their classifications.,,,,,,,,,,
```
maxcheremisin, line 1: am i missing something or does this csv file lack column names?

docs/organization/taxonomy-transform/examples/example-transform.js
mayakarabula: ah yes, I have changed the text to not copy directly from the customer's provided example

AdrianLThomas:
I think the PR has the wrong ticket btw :-)

test/integration/cmds/organization/taxonomy-transform.test.ts
```diff
@@ -0,0 +1,79 @@
...
+          'Usage: contentful organization taxonomy-transform'
+        )
+      })
+      .end(done)
+  })
+
+  test('should return transformed concepts / concepts scheme', done => {
+    app()
+      .run(
+        `${cmd} --organization-id ${organizationId} -t ../test/integration/cmds/organization/example-load.js --save-file false --space-id 123`
```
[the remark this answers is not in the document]
mayakarabula: you mean with the space id? I actually should remove it

test/integration/cmds/organization/example.csv
```diff
@@ -0,0 +1,10 @@
+Animal-Group,Sub-Category,Animal-Name,Skills,Knowledge,Opportunities
```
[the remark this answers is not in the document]
mayakarabula: so one is in public examples and the other in tests, they represent two different ways of writing the CSVs, so just for example reference

test/integration/cmds/organization/taxonomy-transform.test.ts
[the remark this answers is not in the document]
AdrianLThomas: Ah apologies I was just misreading some of the command - but yes it looks like space ID shouldn't be there either?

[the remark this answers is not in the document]
mayakarabula: yeah space id was due to a default constraint I forgot to remove, but now it's cleaned up

AdrianLThomas:
I noticed if we don't pass a `--management-token` then we simply get a 404, should we add validation for that?

lib/cmds/organization_cmds/utils/taxonomy.ts
```diff
@@ -0,0 +1,128 @@
...
+    const existingConcept = this.concepts.existing.find(
+      concept => concept.toJson().id === id
+    )
+    const addedConcept = this.concepts.upsert.find(
+      concept => concept.toJson().id === id
+    )
+
+    if (!existingConcept || !addedConcept) {
+      return null
+    }
```
jjolton-contentful, line 85: curious about this bit of logic -  if either `existingConcept` or `addedConcept` is falsy, the method returns `null`, so could this unintentionally prevent the method from returning a concept even though it exists in one of the arrays?

wondering if it should instead be 
```suggestion
    if (!existingConcept && !addedConcept) {
      return null
    }
```
but there might be a desired behavior here I'm missing

lib/cmds/organization_cmds/utils/taxonomy.ts
```diff
@@ -0,0 +1,128 @@
...
+
+    if (existingConcept) {
+      this.concepts.upsert.push(existingConcept)
```
jjolton-contentful, line 88: wanted to ask, would there be any value in confirming whether `existingConcept` is already in `this.concepts.upsert` before pushing to prevent duplicates?

lib/cmds/organization_cmds/utils/taxonomy.ts
mayakarabula: oh, you are totally right

lib/cmds/organization_cmds/taxonomy-transform.ts
```diff
@@ -0,0 +1,201 @@
...
+
+module.exports.command = 'taxonomy-transform'
+
+module.exports.desc =
+  'transform taxonomy from external format to contentful json format'
+
+module.exports.builder = (yargs: Argv) => {
+  return yargs
+    .usage('Usage: contentful organization taxonomy-transform')
+    .option('management-token', {
```
mayakarabula, line 23: do we need it?

lib/cmds/organization_cmds/taxonomy-transform.ts
```diff
@@ -0,0 +1,201 @@
...
+      type: 'string',
+      demandOption: true
+    })
+    .option('header', {
+      alias: 'H',
+      type: 'string',
+      describe: 'Pass an additional HTTP Header'
+    })
+    .option('output-file', {
+      alias: 'o',
```
mayakarabula, line 40: should we keep capitalization consistent?

lib/cmds/organization_cmds/taxonomy-transform.ts
```diff
@@ -0,0 +1,201 @@
...
+      type: 'string',
+      describe:
+        'Output file. It defaults to ./data/<timestamp>-<organization-id>-transformed.json'
+    })
+    .option('transform-file', {
+      alias: 't',
+      describe: 'File used to transform the taxonomy data',
```
mayakarabula, line 47: be more explicit about this being a js script not csv, maybe transform-script

lib/cmds/organization_cmds/utils/taxonomy.ts
```diff
@@ -0,0 +1,128 @@
...
+    init: Partial<Omit<CreateConceptSchemeProps, 'id'>> & {
+      prefLabel: CreateConceptSchemeProps['prefLabel']
+    }
+  ) {
+    if (
+      this.conceptSchemes.existing.find(
+        conceptScheme => conceptScheme.toJson().id === id
+      )
+    ) {
+      return null
```
mayakarabula, line 105: handle the upsert

lib/cmds/organization_cmds/taxonomy-transform.ts
mayakarabula: we do, otherwise we can't pass token

test/integration/cmds/organization/__snapshots__/taxonomy-transform.test.ts.snap
```diff
@@ -0,0 +1,345 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
momenzadehvida, line 1: do we need this snap test?

mayakarabula: I think it shows us that the different methods work as we expect

momenzadehvida (approved):

jjolton-contentful (approved):
❤️

PR #2937, opened 2025-03-10
taxonomy export csv

momenzadehvida:
Adds a new command to export taxonomy as a .csv file. The CSV should include clearly labeled, easy-to-read column titles (id, PrefLabel).

As a user, I should be able to run contentful organization taxonomy-export --organization-id --format="csv" and get the export of csv like this.

[taxonomy-csv-export.csv](https://github.com/user-attachments/files/19164277/taxonomy-csv-export.csv)

### DIFF
--- a/lib/cmds/organization_cmds/taxonomy-export.ts
+++ b/lib/cmds/organization_cmds/taxonomy-export.ts
@@ -1,6 +1,6 @@
 import Listr from 'listr'
 import { noop } from 'lodash'
-import path from 'path'
+import path, { format } from 'path'
 import type { Argv } from 'yargs'
 import { handleAsyncError as handle } from '../../utils/async'
 import { createPlainClient } from '../../utils/contentful-clients'
@@ -14,6 +14,11 @@ import { buildTree } from './taxonomy/buildTree'
 
 module.exports.command = 'taxonomy-export'
 module.exports.desc = 'export your taxonomy as a csv file'
+
+enum ExportTaxonomyFormat {
+  JSON = 'json',
+  CSV = 'csv'
+}
 
 module.exports.builder = (yargs: Argv) => {
   return yargs
@@ -54,8 +59,9 @@ module.exports.builder = (yargs: Argv) => {
     .option('format', {
       alias: 'F',
       type: 'string',
-      describe: 'Suppress any log output',
-      default: 'json'
+      describe: 'format of the export file',
+      default: 'json',
+      choices: Object.values(ExportTaxonomyFormat)
     })
     .epilog(
       [
--- a/lib/cmds/organization_cmds/taxonomy-export.ts
+++ b/lib/cmds/organization_cmds/taxonomy-export.ts
@@ -1,0 +1,236 @@
+import Listr from 'listr'
+import { noop } from 'lodash'
+import path from 'path'
+import type { Argv } from 'yargs'
+import { handleAsyncError as handle } from '../../utils/async'
+import { createPlainClient } from '../../utils/contentful-clients'
+import { copyright } from '../../utils/copyright'
+import { cursorPaginate } from '../../utils/cursor-pagninate'
+import { ensureDir, getPath, writeFileP } from '../../utils/fs'
+import { getHeadersFromOption } from '../../utils/headers'
+import { success, log } from '../../utils/log'
+import { ConceptProps, ConceptSchemeProps } from 'contentful-management'
+import { buildTree } from './taxonomy/buildTree'
+
+module.exports.command = 'taxonomy-export'
+module.exports.desc = 'export your taxonomy as a csv file'
+
+module.exports.builder = (yargs: Argv) => {
+  return yargs
+    .usage('Usage: contentful organization export')
+    .option('management-token', {
+      alias: 'mt',
+      describe: 'Contentful management API token',
+      type: 'string'
+    })
+    .option('organization-id', {
+      alias: 'oid',
+      describe: 'ID of Organization with source data',
+      type: 'string',
+      demandOption: true
+    })
+    .option('header', {
+      alias: 'H',
+      type: 'string',
+      describe: 'Pass an additional HTTP Header'
+    })
+    .option('output-file', {
+      alias: 'o',
+      type: 'string',
+      describe:
+        'Output file. It defaults to ./data/<timestamp>-<organization-id>.csv'
+    })
+    .option('silent', {
+      alias: 'S',
+      type: 'boolean',
+      describe: 'Suppress any log output',
+      default: false
+    })
+    .option('save-file', {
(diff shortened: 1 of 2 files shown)


lib/cmds/organization_cmds/taxonomy-export.ts
```diff
@@ -0,0 +1,244 @@
...
+              }
+            }
+          ])
+        }
+      }
+    ],
+    { renderer: silent ? 'silent' : 'default' }
+  )
+
+  const result = await tasks.run({
```
BobHemphill76, line 172: any chance we can get tests around result processing?

lib/cmds/organization_cmds/taxonomy/buildTree.ts
```diff
@@ -0,0 +1,46 @@
+type ItemId = string | null
+export type Nested<T> = { children: Array<Nested<T>> } & {
+  [K in keyof T]: T[K]
+}
+
+export function buildTree<T, R extends Nested<T>>(
```
BobHemphill76, line 6: can we test this?

lib/cmds/organization_cmds/taxonomy-export.ts
```diff
@@ -0,0 +1,244 @@
...
+      taxonomyTree[broader.sys.id].children.push(taxonomyTree[concept.sys.id])
+    })
+  })
+
+  if (format === 'csv') {
+    const csvExportTitle: string[] = [
+      'Concept scheme id',
+      'Concept scheme preferred label'
+    ]
+    result.taxonomy.conceptSchemes.map((scheme: ConceptSchemeProps) => {
```
BobHemphill76, line 191: doesn't look like this map is returning anything and can probably be changed to forEach

lib/cmds/organization_cmds/taxonomy-export.ts
```diff
@@ -0,0 +1,244 @@
...
+      const conceptInSchemeIds = scheme.concepts.map(c => c.sys.id)
+
+      const concepts = result.taxonomy.concepts.filter((c: ConceptProps) =>
```
BobHemphill76, line 194: This ends up being O(N*M) with N being len(taxony.concepts) and M being len(scheme.concepts). Do concepts have an upper bound?  If not, it might be better to turn the result of 192 into a set / map rather than a list, so we can reduce this complexity to O(N)...especially given we're already in a map function.

lib/cmds/organization_cmds/taxonomy-export.ts
```diff
@@ -0,0 +1,244 @@
...
+        conceptInSchemeIds.includes(c.sys.id)
+      )
+
+      const tree = buildTree(concepts, {
+        getItemId: (c: ConceptProps) => c.sys.id,
+        getParentIds: c => {
+          const parentIds: string[] = []
+          for (const b of c.broader) {
+            if (conceptInSchemeIds.includes(b.sys.id)) {
```
BobHemphill76, line 203: related to comment on 192.  I'm assuming getParentIds is called in some kind of loop?  If so, I would def convert conceptInSchemeIds into a more performant lookup data structure.

lib/cmds/organization_cmds/taxonomy-export.ts
```diff
@@ -0,0 +1,244 @@
...
+              parentIds.push(b.sys.id)
+            }
+          }
+          return parentIds.length ? parentIds : [null]
+        }
+      })
+
+      csvExports.push(`${scheme.sys.id},${scheme.prefLabel['en-US']},`)
```
BobHemphill76, line 211: is it intended that the csvExports has multiple rows per taxonomy.conceptSchemes?

lib/cmds/organization_cmds/taxonomy-export.ts
```diff
@@ -0,0 +1,244 @@
...
+      await writeFileP(
+        outputTarget,
+        [...[csvExportTitle.join(',')], ...csvExports].join('\n')
+      )
+    } else {
+      log(JSON.stringify(result, null, 2))
+    }
+  } else {
+    if (saveFile) {
+      await writeFileP(outputTarget, JSON.stringify(result, null, 2))
```
BobHemphill76, line 231: This output type seems radically different than the csv output type.  ie: multi line / scheme with flattened depth headers for the csv vs { taxonomy: { concepts: [], conceptSchemes: [] } } for the json.  Was this intentional?

momenzadehvida: For exporting to JSON, we follow the same process as with all other entities. However, for the CSV format, this is the specific format requested by our customer. They need to be able to understand the hierarchy by reviewing the CSV, where each parent is followed by its children in subsequent rows. It should use two columns to represent the hierarchy and include only the id and prefLabel fields.

lib/cmds/organization_cmds/taxonomy-export.ts
momenzadehvida: For each scheme we will have just one row and all concepts will be followed by its concepts in subsequent rows.

lib/cmds/organization_cmds/taxonomy-export.ts
momenzadehvida: yes, We have upper limits for the number of concepts and concept schemes within a single organization.

lib/cmds/organization_cmds/taxonomy/buildTree.ts
momenzadehvida: I added unit-tests for it.

lib/cmds/organization_cmds/taxonomy-export.ts
```diff
@@ -0,0 +1,244 @@
...
+    feature: 'organization-export',
+    headers: getHeadersFromOption(header),
+    throttle: 8,
+    logHandler: noop
+  })
+
+  let maxDepth = 1
+  const csvExports: string[] = []
+
+  const findAllHeirachy = (
```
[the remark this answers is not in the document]
momenzadehvida: I extracted the function and  added unit-tests for it.

lib/cmds/organization_cmds/taxonomy-export.ts
momenzadehvida: The maximum number of schemes allowed in one organization is currently 20, which is not a high value, and we do not anticipate increasing it.

lib/cmds/organization_cmds/taxonomy-export.ts
momenzadehvida: I have already covered the steps in my tests. Are there any other tests you would recommend adding?

mayakarabula (approved):

mayakarabula (approved):

michaelpineirocontentful (approved):

PR #2944, opened 2025-03-18
fix: allow --upload-assets and --assets-directory import flags

joelmiles:
The contentful-import library supports the [--uploadAssets](https://github.com/contentful/contentful-import?tab=readme-ov-file#uploadassets-boolean-default-false) and [--assetsDirectory](https://github.com/contentful/contentful-import?tab=readme-ov-file#uploadassets-boolean-default-false) flags. This change allows these flags to be propagated from the CLI to the import library


joelmiles:
Is anything else needed here before a merge? We'd really like to use this feature and not rely on home-built binaries.

ethan-ozelius-contentful:
Hello @joelmiles, thanks for raising this PR.  You are 100% correct that this functionality belongs in the contentful-cli tool.  A PR was merged to the CLI tool last month that added these arguments.  Please refer to PR/new version.

[PR](https://github.com/contentful/contentful-cli/pull/2977)
[NPM package with fix](https://www.npmjs.com/package/contentful-cli/v/3.8.8)

PR #2977, opened 2025-04-14
[DX-28] fix(import-tool): add --upload-assets and --assets-directory params to contentful-cli

ethan-ozelius-contentful:
## Summary
These 2 params are already supported by contentful-import, but were not being passed by contentful-cli.  For what it's worth, these 2 params were actually working as long as the param was hyphenated, instead of camelCased. i.e.  

```
// currently works
  --upload-assets
  --assets-directory

// currently does not work, results in "unknown argument", which is fixed by this PR
  --uploadAssets
  --assetsDirectory
```

 But now that I have added the `yargs` definition, camelCase is supported, and the param info is displayed with running `$ contentful space/organization --help` 

### Before
```
$ npx contentful space import\
  --space-id <mySpaceId>\
  --environment-id testing\
  --management-token "$CONTENTFUL_PRODUCTION_ACCESS_TOKEN"\
  --uploadAssets\
  --assetsDirectory ./images.ctfassets.net\
  --content-file ./exported/contentful-export-<mySpaceId>-master-2025-04-14T11-58-46.json

Usage: contentful space import --content-file <file>

Options:
  ...
  --config                   An optional configuration JSON file containing all
                             the options for a single run

Copyright 2019 Contentful
Unknown arguments: uploadAssets, assetsDirectory. // ka-boom
```

### After
```
$ npx contentful space import\
  --space-id <mySpaceId>\
  --environment-id testing\
  --management-token "$CONTENTFUL_PRODUCTION_ACCESS_TOKEN"\
  --uploadAssets\
  --assetsDirectory ./images.ctfassets.net\
  --content-file ./exported/contentful-export-<mySpaceId>-master-2025-04-14T11-58-46.json

┌──────────────────────────────────────────────────┐
│ The following entities are going to be imported: │
├────────────────────────────────┬─────────────────┤
│ Content Types                  │ 1               │
├────────────────────────────────┼─────────────────┤
│ Tags                           │ 150             │
├────────────────────────────────┼─────────────────┤
│ Editor Interfaces              │ 1               │
├────────────────────────────────┼─────────────────┤
│ Entries                        │ 0               │
├────────────────────────────────┼─────────────────┤
│ Assets                         │ 2               │
├────────────────────────────────┼─────────────────┤
│ Locales                        │ 102             │
├────────────────────────────────┼─────────────────┤
│ Webhooks                       │ 1               │
└────────────────────────────────┴─────────────────┘
```


elylucas (approved):
Looks pretty good to me, though I'd wait for someone more knowledgable on the codebase. I do appreciate the updated readmes!

lib/cmds/organization_cmds/import.ts
```diff
@@ -55,6 +55,18 @@ module.exports.builder = (yargs: Argv) => {
...
+      type: 'boolean',
+      describe:
+        'Upload raw assets from local file system (as opposed to via URL).  If true, --assets-directory must also be provided.',
+      default: false
+    })
+    .option('assets-directory', {
+      type: 'string',
+      describe:
+        'Relative path to exported asset files to be uploaded (as opposed to via URL).',
+      default: false
```
whitelisab, line 68: Maybe remove `default: false` for this option since it's a string? We probably don't want to set this option's default to `false`.

docs/organization/import/README.md
```diff
@@ -0,0 +1,54 @@
...
+                                        [string] [default: "api.contentful.com"]
+  --proxy                    Proxy configuration in HTTP auth format:
+                             [http|https]://host:port or
+                             [http|https]://user:password@host:port     [string]
+  --timeout                  Timeout in milliseconds for API calls
+                                                       [number] [default: 20000]
+  --retry-limit              How many times to retry before an operation fails
+                                                          [number] [default: 10]
+  --header, -H               Pass an additional HTTP Header             [string]
+  --upload-assets            Upload exported files instead of pointing to an
```
whitelisab, line 34: We talked about how maybe we could adopt the language that is currently in the `README` for the contentful-import tool, which talks about how these options are needed when using the `downloadAssets` option of the export tool. There's several references/instructions in this PR that you can take a look at and update for consistency

lib/cmds/organization_cmds/import.ts
whitelisab, line 1: Let's check to make sure these options make sense for the organization import command

whitelisab (commented):
Awesome stuff! ⭐  Just a few small comments for you to review

CONTRIBUTING.md
```diff
@@ -93,6 +93,4 @@ Both of these test environments are setup to deal with Babel and code transpilin
 
 # Other tasks
```
ethan-ozelius-contentful, line 95: remove npm scripts that not longer exist.

docs/organization/import/README.md
ethan-ozelius-contentful, line 1: This is a net new readme.  The `$ contentful organization import ...` command has existed for a while, but looks like a readme was never created for it.  Interestingly, 

1. importing/exporting at the organization level only imports/exports **Taxonomy** data.
2. It doesn't seem like this functionality existed in the original `contentful-export` tool, but I could be incorrect on that.

docs/space/import/README.md
```diff
@@ -20,9 +20,12 @@ Options:
   --skip-locales             Skip importing locales   [boolean] [default: false]
   --skip-content-publishing  Skips content publishing. Creates content but does
                              not publish it           [boolean] [default: false]
```
ethan-ozelius-contentful, line 22: `--skip-content-updates` and `--skip-asset-updates` already exist and are unrelated to this PR.  This readme was just never updated.

same story with --host.

lib/cmds/organization_cmds/import.ts
ethan-ozelius-contentful: Good call, paired on this and investigated organizations imports a bit on my own.  Entries/Assets and any other **space** scoped entities are not applicable to **organization** imports/exports.  So I have removed the functional changes from this file, and just added the required params to the usage hint.

For what it's worth, as mentioned in a previous comment Organization imports/exports deal _only_ with Taxonomies, and I don't think was part of the original contentful-import and contentful-export tools.

lib/cmds/organization_cmds/import.ts
ethan-ozelius-contentful: good catch, but no longer applicable since I reverted these changes to this file.

docs/organization/import/README.md
ethan-ozelius-contentful: updated to use original copy from `contentful-import` [tool](https://github.com/contentful/contentful-import/blob/main/README.md).

```md
#### `uploadAssets` [boolean] [default: false]

Upload local asset files downloaded via the [downloadAssets](https://github.com/contentful/contentful-export#downloadassets-boolean) option of the export. Requires `assetsDirectory`

#### `assetsDirectory` [string]

Path to a directory with an asset export made using the [downloadAssets](https://github.com/contentful/contentful-export#downloadassets-boolean) option of the export. Requires `uploadAssets`

```

docs/organization/import/README.md
```diff
@@ -0,0 +1,31 @@
...
+### Example
+
+```sh
+contentful organization import --content-file <path/to/content_file.json> --organization-id <organization-id>
+```
+
+## Limitations
+
+- This tool only imports **Taxonomy** entities. Any other Organization level entities will not be imported.
+- For Space scoped entites (Entries, Assets, Apps etc) see [space command](https://github.com/contentful/contentful-cli/blob/main/docs/space/import/README.md)
```
whitelisab, line 31: Small typo, `entites` should be `entities`

```suggestion
- For Space scoped entities (Entries, Assets, Apps etc) see [space command](https://github.com/contentful/contentful-cli/blob/main/docs/space/import/README.md)
```

docs/space/import/README.md
```diff
@@ -20,9 +20,12 @@ Options:
...
+  --skip-content-updates     Skips updating existing content, only creates new
+                             entries.                 [boolean] [default: false]
+  --skip-asset-updates       Skips updating existing assets, only creates new
+                             assets.                  [boolean] [default: false]
   --error-log-file           Full path to the error log file            [string]
-  --host                     Management API host
-                                        [string] [default: "api.contentful.com"]
```
whitelisab, line 25: Is `api.contentful.com` still the default for host? Should we keep that noted in the README?

whitelisab (approved):
I left two small comments on `README` content, but otherwise looks great, approving!

docs/space/import/README.md
ethan-ozelius-contentful: No, the defaults have been removed from the space import params, see this [PR](https://github.com/contentful/contentful-cli/pull/2169/files) for more context.  But the gist is (copy'd/pasted from PR description):

> With the code in config construction middleware, we are always assigning a value of the host to the one from args if it is present, see here: https://github.com/contentful/contentful-cli/blob/fix/PHX-2732-export-import-commands-host/lib/utils/middlewares.js#L50

> having a default value causes host args to be always present and thus always overrides the .contentfulrc.json value.

> This means that if we set it to something different eg staging host it will be ignored

Seems like the readme was just never updated.  I'm starting to think that having the output of `$ contentful <command> --help` in the readme isn't a great idea. 🤔

whitelisab (approved):

PR #3005, opened 2025-05-09
Creating an NA <-> EU migration option 

Chaffexd:
I've spent some time adding functionality to allow bi-directional space transfer(s) between NA and EU data centers as this functionality doesn't currently exist. 

How it works:
contentful migrate datacenter --help                                                  ✔    at 18:35:19  

 migrate datacenter

Migrate a space between Contentful data centers

Options:
  --source                   Source data center (na or eu)
                                                [required] [choices: "na", "eu"]
  --target                   Target data center (na or eu)
                                                [required] [choices: "na", "eu"]
  --source-space-id          Source space ID                 [string] [required]
  --target-space-id          Target space ID                 [string] [required]
  --environment-id           Environment ID (e.g. master)
                                                    [string] [default: "master"]
  --source-token             Source CMA token                [string] [required]
  --target-token             Target CMA token                [string] [required]
  --use-verbose-renderer     Display progress in new lines instead of spinner
                                                      [boolean] [default: false]
  --include-taxonomies       Migrate taxonomies (Concepts & Concept Schemes)
                             before importing content [boolean] [default: false]
  --source-org-id            Organization ID where taxonomies should be exported
                             from                                       [string]
  --target-org-id            Organization ID where taxonomies should be imported
                             to                                         [string]

An example:
contentful migrate datacenter \
--source na \
--target eu \
--source-space-id <spaceId> \
--target-space-id <spaceId> \
--source-token <PAT Token> \
--target-token <PAT Token> \
--source-org-id <orgId> \
--target-org-id <orgId> \
--include-taxonomies <boolean> \
--skip-content-publishing <boolean>

I have manually tested this and it works both ways, it allows you to include taxonomies if you use them as otherwise the space import will fail, if you don't use them then you can simple use --include-taxonomies false


lib/cmds/migrate_cmds/datacenter.js
```diff
@@ -0,0 +1,194 @@
...
+    environmentId,
+    useVerboseRenderer
+  } = argv
+
+  const exportDir = path.join(process.cwd(), 'tmp-export')
+  ensureDirExists(exportDir)
+
+  if (argv.includeTaxonomies) {
+    if (!argv.sourceOrgId || !argv.targetOrgId) {
+      throw new Error(
```
ethan-ozelius-contentful, line 40: Optional suggestion, you can use [yargs implies](https://yargs.js.org/docs/#api-reference-impliesx-y) instead of manually validating/throwing.

lib/cmds/migrate_cmds/datacenter.js
```diff
@@ -0,0 +1,194 @@
...
+    )
+
+    const {
+      organizationExport
+    } = require('../../cmds/organization_cmds/export')
+
+    await organizationExport({
+      context: { managementToken: argv.sourceToken },
+      organizationId: argv.sourceOrgId,
+      header: argv.header,
```
ethan-ozelius-contentful, line 61: `header` needs to be a defined property in the yargs right?

ethan-ozelius-contentful:
Hello @Chaffexd, thanks a bunch for investing the time to improve this tool.  I left a couple non-blocking questions/suggestions but otherwise the code looks very clean and well tested.

After discussing with the Developer Experience PM, we have a couple thoughts around the name of the command `migrate-datacenter`.   And while we fully acknowledge that we're bikeshedding here, we have 2 ideas we'd love your thoughts on:

1. instead of using `datacenter`, use `region`.  Contentful doesn't have physical datacenters, and conceptually I defer to the [AWS regional model](https://aws.amazon.com/about-aws/global-infrastructure/regions_az/), where you can pick a `region` like `US East (Northern Virginia) Region` or `Europe (Frankfurt) Region`.  But if you think our customers would understand datacenter more easily, we're open to it.
2. instead of using `migration`, use `sync`?  There already exists the concept of `migration` in the [cli tool](https://github.com/contentful/contentful-cli/blob/main/lib/cmds/space_cmds/migration.js#L86), which is a migration in the sense that data is being mutated.  This PR is more exporting/importing data from different continents, but not making any changes to the underlying data.

Let us know what you think!

Lastly, this PR would really benefit from having a dedicated doc for the new command, similar to [other commands](https://github.com/contentful/contentful-cli/tree/main/docs) have readme with basic descriptions/options/limitations.

lib/cmds/migrate_cmds/datacenter.js
Chaffexd: after review, its actually deadweight so I'm removing it

lib/cmds/migrate_cmds/datacenter.js
```diff
@@ -0,0 +1,194 @@
...
+      host: HOST_MAP[target]
+    })
+
+    logging.success(`${emojic.whiteCheckMark} Taxonomy export complete`)
+
+    console.log(
+      `${emojic.inboxTray} Importing taxonomies into target organization ${argv.targetOrgId}...`
+    )
+
+    const {
```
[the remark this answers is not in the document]
Chaffexd: at the time of writing, I just put them there but moving them to be globally available seems unnecessary

ethan-ozelius-contentful (approved):

PR #3264, opened 2026-04-05
upgrade to inquirer@8.2.7 patches CVE-2025-54798;

benpetty:
## Summary

Upgrade inquirer from `^8.0.0` to `^8.2.7` which patches [CVE-2025-54798](https://github.com/advisories/GHSA-52f5-9888-hmc6)

## Description

`tmp@0.2.3` is vulnerable to an Arbitrary temporary file / directory write via symbolic link dir parameter.  `contentful-cli@3.10.4` requires `tmp@^0.0.29` via a transitive dependency on `external-editor@1.1.1`. Upgrading inquirer fixes this :)

re: 
- https://github.com/SBoudrias/Inquirer.js/issues/1802#issuecomment-3169341642
- https://github.com/SBoudrias/Inquirer.js/pull/1804
- https://github.com/SBoudrias/Inquirer.js/releases/tag/inquirer%408.2.7

PR #3269, opened 2026-04-23
build(deps): bump inquirer to 8.2.7[]

michaelphamcf:
## Summary

inquirer `8.2.7` is already installed, which patched [CVE-2025-54798](https://github.com/advisories/GHSA-52f5-9888-hmc6). This PR raises the version floor to `^8.2.7`.

Original PR: https://github.com/contentful/contentful-cli/pull/3264 

## Todos

- [ ] Implemented feature
- [ ] Feature with pending implementation


package-lock.json
whitelisab, line 1: It looks like package-lock.json is missing the actual resolved entry update; only the top-level packages section (which mirrors package.json's dependency range) is updated in this file. Running npm install should make the other necessary changes to the package-lock.json file

michaelphamcf: Thanks for pointing that out. Looked into it a little bit more, and turns out 8.2.7 is already installed. With that, I've updated the PR description raise the version floor to 8.2.7

whitelisab (approved):

